Issue #10, opened 2018-06-17
Add more references


chris-wood:
https://eprint.iacr.org/2016/144
http://webee.technion.ac.il/~hugo/sphinx.pdf

alxdavids:
https://eprint.iacr.org/2014/650

alxdavids:
will do this

PR #11, opened 2018-06-22
add references

alxdavids:
#10 
(also updated the .txt file)


chris-wood:
I think the hash2curve reference should be: I-D.irtf-cfrg-.... Can you please give that a shot? Also, can you please sort the authors in alphabetical order by surname? (You, Nick, then myself.)

alxdavids:
Okay I've changed the references back, they don't seem to build correctly (I think the ref is missing). Also switched the authors up.

chris-wood:
I fixed the references for you -- all set to go!

PR #12, opened 2018-10-18
Using SWU encoding for hash-to-curve

alxdavids:
Not an expert with hashing to curves but it seems like the SWU encoding seems to work for P-256, so shall we go with that?  We already have noted with privacy pass that we should look to move towards this method of encoding so it seems to fit well.


alxdavids:
Ooops sorry, yeah you're right. I'll change them.

alxdavids:
I've updated now, also using Icart for P384 as per the h2c draft.

PR #23, opened 2019-03-10
Add notes on additional key consistency constraints.

e271828-:
Covers some of the points we went over in person last month, and corrects a few typos.


alxdavids:
Thanks for the suggested changes. We're currently working from draft-sullivan-cfrg-voprf.md for the latest draft, would it be possible to make the changes to this file instead? Thanks!

PR #25, opened 2019-03-11
Add notes on additional key consistency constraints.

e271828-:
Covers some of the points we went over in person last month, and corrects a few typos.

(On the correct document this time.)

### DIFF
--- a/draft-sullivan-cfrg-voprf.md
+++ b/draft-sullivan-cfrg-voprf.md
@@ -1010,10 +1010,19 @@ CURVE25519 and CURVE448.
 
 DLEQ proofs are essential to the protocol to allow V to check that P's
 designated private key was used in the computation. A side effect of this
-property is that it prevents P from using unique key for select verifiers as a
-way of "tagging" them. If all verifiers expect use of a certain private key,
-e.g., by locating P's public key key published from a trusted registry, then P
+property is that it prevents P from using a unique key for select verifiers as
+a way of "tagging" them. If all verifiers expect use of a certain private key,
+e.g., by locating P's public key published from a trusted registry, then P
 cannot present unique keys to an individual verifier.
+
+For this side effect to hold, P must also be prevented from using other 
+unblinding techniques that can be achieved via manipulation of P's public key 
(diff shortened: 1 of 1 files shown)


draft-sullivan-cfrg-voprf.md
```diff
@@ -1010,11 +1010,20 @@ CURVE25519 and CURVE448.
...
-e.g., by locating P's public key key published from a trusted registry, then P
+property is that it prevents P from using a unique key for select verifiers as
+a way of "tagging" them. If all verifiers expect use of a certain private key,
+e.g., by locating P's public key published from a trusted registry, then P
 cannot present unique keys to an individual verifier.
 
+For this side effect to hold, P must also be prevented from using other 
+unblinding techniques that can be achieved via manipulation of P's public key 
+within a trusted registry. For example, if P is able to rotate P's published 
+public key at short time intervals, P may be able to assign a likely identity
```
chris-wood: Hmm, I don't quite understand this example. Could we provide some more details?

alxdavids: I believe this is just referring to the fact that having short key cycles can lead to a similar deanonymising effect as using individual keys for each client V.

Perhaps we could reword to the following:

> For this side effect to hold, P must also be prevented from using other techniques that can be achieved via manipulation of P's public key within a trusted registry. For example, if P's public key is rotated too frequently then this may stratify the user base into small anonymity groups (those with VOPRF_Sign outputs taken from a given key epoch). In this case, it may become practical to link VOPRF sessions for a given user and thus compromises their privacy.

chris-wood:
@e271828- can you please resolve the conflicts so we can merge?

draft-sullivan-cfrg-voprf.md
e271828-: This is correct. I've accepted @alxdavids wording and resubmitted in https://github.com/chris-wood/draft-sullivan-cfrg-oprf/pull/26 to keep the diff clean.

e271828-:
Closing in favor of https://github.com/chris-wood/draft-sullivan-cfrg-oprf/pull/26

PR #42, opened 2019-07-05
Streamline and correct OPRF security properties

### LINKED ISSUE #32: Clean up security properties for OPRF
The pseudorandomness of the function dictates that it is unlinkable and non-malleable, so we can remove these properties. There are some other things that Hugo brought up:

```
   o  The output distribution of y = F(k, x) is indistinguishable from
      the uniform distribution in the domain of the function F.

> domain -> range

> It is more general and "more correct" to define PRF as in the introduction
> "F is pseudorandom if the output y = F(k, x) is indistinguishable from
> uniformly sampling any element in F's range for random choice of k."
> From this all the properties you list follow (including the "additional ones")
```

```
   Additionally, we require the following additional properties:

> additional twice
```

```
   o  Oblivious: P must learn nothing about V's input, and V must learn
      nothing about P's private key.

> Oblivious should include also "P learns nothing about the output of the
> function". Formally, this is redundant but it is a very important consequence
> of the definition and it is better to say it explicitly.
```

```
   o  Verifiable: V must only complete execution of the protocol if it
      can successfully assert that P used its secret key k.

> "used its key" is a bit too vague. Maybe: if it can successfully assert that
> the OPRF output V computes is correct with respect to the OPRF key k held by
> P.
```

```
   In practice, the notion of verifiability requires that P commits to
   the key k before the actual protocol execution takes place.  Then V
   verifies that P has used k in the protocol using this commitment.

> We could add: Below we will refer to this commitment to k as a "public key".
```

alxdavids:
Fixes #32

### DIFF
--- a/draft-sullivan-cfrg-oprf.md
+++ b/draft-sullivan-cfrg-oprf.md
@@ -293,28 +293,34 @@ knowledge of k. This is discussed more in the following section.
 The security properties of an OPRF protocol with functionality y = F(k, x)
 include those of a standard PRF. Specifically:
 
-- Given value x, it is infeasible to compute y = F(k, x) without knowledge of k.
-- The output distribution of y = F(k, x) is indistinguishable from the uniform
-  distribution in the domain of the function F.
+- Pseudorandomness: F is pseudorandom if the output y = F(k,x) on any input x
+  is indistinguishable from uniformly sampling any element in F's range, for a
+  random sampling of k.
 
-Additionally, we require the following additional properties:
+In other words, for an adversary that can pick inputs x from the domain of F and
+can evaluate F on (k,x) (without knowledge of randomly sampled k), then the
+output distribution F(k,x) is indistinguishable from the uniform distribution in
+the range of F.
 
-- Non-malleable: Given (x, y = F(k, x)), V must not be able to generate (x', y')
-  where x' != x and y' = F(k, x').
-- Oblivious: P must learn nothing about V's input, and V must learn nothing
-  about P's private key.
-- Unlinkable: If V reveals x to P, P cannot link x to the protocol instance in
-  which y = F(k, x) was computed.
+The OPRF protocol must also satisfy the following property:
+
+- Oblivious: P must learn nothing about V's input or the output of the function.
+  In addition, V must learn nothing about P's private key.
 
 Optionally, for any protocol that satisfies the above properties, there is an
 additional security property:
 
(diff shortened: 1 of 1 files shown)


draft-sullivan-cfrg-oprf.md
```diff
@@ -293,28 +293,34 @@ knowledge of k. This is discussed more in the following section.
...
+  is indistinguishable from uniformly sampling any element in F's range, for a
+  random sampling of k.
 
-Additionally, we require the following additional properties:
+In other words, for an adversary that can pick inputs x from the domain of F and
+can evaluate F on (k,x) (without knowledge of randomly sampled k), then the
+output distribution F(k,x) is indistinguishable from the uniform distribution in
+the range of F.
 
-- Non-malleable: Given (x, y = F(k, x)), V must not be able to generate (x', y')
```
chris-wood: Why did we remove this property?

draft-sullivan-cfrg-oprf.md
```diff
@@ -293,28 +293,34 @@ knowledge of k. This is discussed more in the following section.
...
-  where x' != x and y' = F(k, x').
-- Oblivious: P must learn nothing about V's input, and V must learn nothing
-  about P's private key.
-- Unlinkable: If V reveals x to P, P cannot link x to the protocol instance in
```
chris-wood: And this one?

alxdavids:
The reason that I removed these definitions is because they're implied by pseudorandomness. Any function that is malleable cannot be pseudorandom, because it would be distinguishable from some random function (which is not malleable with high probability). The same with unlinkability, if you have obliviousness and pseudorandomness then you shouldn't be able to invert the function.

If you would prefer I could add some text that describes that (based on pseudoranomness and unlinkability) then we have unlinkability and non-malleability. But, I agree with Hugo that listing these properties separately from pseudorandomness is unnecessary.

chris-wood:
> If you would prefer I could add some text that describes that (based on pseudoranomness and unlinkability) then we have unlinkability and non-malleability.

Yeah, I think adding even parenthetical text that says these properties hold is useful. Can we do that?

alxdavids:
> > If you would prefer I could add some text that describes that (based on pseudoranomness and unlinkability) then we have unlinkability and non-malleability.
> 
> Yeah, I think adding even parenthetical text that says these properties hold is useful. Can we do that?

Sure, will do!

alxdavids:
@chris-wood let me know what you think about the new text that I have added.

PR #44, opened 2019-07-05
Replace PRNG with additional RO function

### LINKED ISSUE #36: Replace PRNG with RO for batched proof
As Hugo notes:
```
 2. d1,...dn <- PRNG(seed,salt,info,n)

> PRNGs inputs are just a seed, usually a secret one. There are no guarantees
> about the pseudorandomness of the output for a public seed. Additional items
> such as salt, info, n, are not accommodated at all in a PRNG. You could put
> these under H_4 or use a (variable-length output) PRF instead of PRNG.
> But also in the PRF case, non-secret key does not guarantee pseudorandomness.
> What you seem to need is a RO which on inputs G,Y,[Mi,Zi],salt,info,n outputs 
> d1,...dn (is there a proof that batching is secure when both parties derive d_i
> using a RO?)

> Btw, independently of the above issues, if you let the prover choose the salt,
> this can help him (at least in theory) to cheat - i.e., finding a set of
> "misleading"  d's.
```
I think it will be easiest to introduce an additional RO `H_4` that handles the generation of `d1,...,d4` here.

We would also need to clarify the best instantiations for doing this, it may be that we stick with SHAKE-SHA256 and HKDF-Expand-SHA256 (this should also be clarified as we currently just say HKDF-256). The notion of the `salt` should be replaced with a domain separation label.

alxdavids:
Fixes #36
- Usage of PRNG is unnecessary.
- We now instantiate with a generic RO
- Specific ciphersuites are unchanged
- HKDF -> HKDF_Expand

### DIFF
--- a/draft-sullivan-cfrg-oprf.md
+++ b/draft-sullivan-cfrg-oprf.md
@@ -928,11 +928,11 @@ same applies to the algorithm VOPRF_Sign. The same applies for replacing
 DLEQ_Verify with Batched_DLEQ_Verify when V verifies the response from P and
 during the algorithm VOPRF_Verify.
 
-## Random oracle instantions for proofs
+## Random oracle instantiations for proofs
 
 We can instantiate the random oracle function H_4 using the same hash function
 that is used for H_1,H_2,H_3. For H_5, we can also use a similar instantiation,
-or we can use a stream-based output generator. For example, for groups with an
+or we can use a variable-length output generator. For example, for groups with an
 order of 256-bit, valid instantiations include functions such as SHAKE-256
 {{SHAKE}} or HKDF-Expand-SHA256 {{RFC5869}}.
 
@@ -941,7 +941,7 @@ used, we note that the outputs of H_5 (d1,...,dn) must be smaller than this
 order. If any di that is sampled is larger than then order, then we should
 resample until a di' is sampled that is valid.
 
-In these cases, the iterating integer i is increased to i' until such di' is
+In these cases, the iterating integer i is increased monotonically to i' until such di' is
 sampled. When sampling the next value d(i+1), the counter i+1 is started at
 i'+1.
 
--- a/draft-sullivan-cfrg-oprf.md
+++ b/draft-sullivan-cfrg-oprf.md
@@ -934,7 +934,7 @@ We can instantiate the random oracle function H_4 using the same hash function
 that is used for H_1,H_2,H_3. For H_5, we can also use a similar instantiation,
 or we can use a stream-based output generator. For example, for groups with an
 order of 256-bit, valid instantiations include functions such as SHAKE-256
-{{SHAKE}} or HKDF_Expand-SHA256 {{RFC5869}}.
+{{SHAKE}} or HKDF-Expand-SHA256 {{RFC5869}}.
 
 In addition if a function with larger output than the order of the base field is
 used, we note that the outputs of H_5 (d1,...,dn) must be smaller than this
@@ -944,6 +944,8 @@ resample until a di' is sampled that is valid.
 In these cases, the iterating integer i is increased to i' until such di' is
 sampled. When sampling the next value d(i+1), the counter i+1 is started at
 i'+1.
+
+TODO: Give a more detailed specification of this construction.
 
 # Supported ciphersuites {#ciphersuites}
 
@@ -963,7 +965,7 @@ ciphersuites demonstrating 128 bits of security.
 - H_2: SHA256
 - H_3: SHA256
 - H_4: SHA256
-- H_5: HKDF_Expand-SHA256
+- H_5: HKDF-Expand-SHA256
 
 ## ECVOPRF-RISTRETTO-HKDF-SHA512-Elligator2:
 
@@ -973,7 +975,7 @@ ciphersuites demonstrating 128 bits of security.
 - H_2: SHA512
 - H_3: SHA512
 - H_4: SHA512
-- H_5: HKDF_Expand-SHA512
+- H_5: HKDF-Expand-SHA512
 
 In the case of Ristretto, internal point representations are represented by
 Ed25519 {{RFC7748}} points. As a result, we can use the same hash-to-curve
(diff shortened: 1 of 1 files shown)


draft-sullivan-cfrg-oprf.md
```diff
@@ -923,22 +922,28 @@ Steps:
...
-multiple blinded PRF inputs [Mi] for i in 1...n. Then P computes the array [Zi]
-and replaces DLEQ_Generate with Batched_DLEQ_Generate over these arrays. The
+multiple blinded PRF inputs [ Mi ] for i in 1...n. Then P computes the array [Zi
+] and replaces DLEQ_Generate with Batched_DLEQ_Generate over these arrays. The
 same applies to the algorithm VOPRF_Sign. The same applies for replacing
 DLEQ_Verify with Batched_DLEQ_Verify when V verifies the response from P and
 during the algorithm VOPRF_Verify.
 
-## PRNG and resampling
+## Random oracle instantions for proofs
```
chris-wood:
```suggestion
## Random oracle instantiations for proofs
```

draft-sullivan-cfrg-oprf.md
```diff
@@ -923,22 +922,28 @@ Steps:
...
 
-Any function that satisfies the security properties of a pseudorandom number
-generator can be used for computing the batched DLEQ proof. For example,
-SHAKE-256 {{SHAKE}} or HKDF-SHA256 {{RFC5869}} would be reasonable choices for
-groups that have an order of 256 bits.
+We can instantiate the random oracle function H_4 using the same hash function
+that is used for H_1,H_2,H_3. For H_5, we can also use a similar instantiation,
+or we can use a stream-based output generator. For example, for groups with an
```
chris-wood:
```suggestion
or we can use a variable-length output generator. For example, for groups with an
```

draft-sullivan-cfrg-oprf.md
```diff
@@ -923,22 +922,28 @@ Steps:
...
+order of 256-bit, valid instantiations include functions such as SHAKE-256
+{{SHAKE}} or HKDF_Expand-SHA256 {{RFC5869}}.
```
chris-wood: Is one preferred over the other? Why? (We should probably include some rationale text here.)

draft-sullivan-cfrg-oprf.md
```diff
@@ -923,22 +922,28 @@ Steps:
...
 
-We note that the PRNG outputs d1,...,dn must be smaller than the order of the
-group/curve that is being used. Resampling can be achieved by increasing the
-value of the iterator that is used in the info field of the PRNG input.
+In addition if a function with larger output than the order of the base field is
+used, we note that the outputs of H_5 (d1,...,dn) must be smaller than this
+order. If any di that is sampled is larger than then order, then we should
+resample until a di' is sampled that is valid.
+
+In these cases, the iterating integer i is increased to i' until such di' is
```
chris-wood:
```suggestion
In these cases, the iterating integer i is increased monotonically to i' until such di' is
```

chris-wood: In general, this is underspecified. We should leave a TODO here at least to write this out.

draft-sullivan-cfrg-oprf.md
```diff
@@ -958,7 +963,7 @@ ciphersuites demonstrating 128 bits of security.
 - H_2: SHA256
 - H_3: SHA256
 - H_4: SHA256
-- PRNG: HKDF-SHA256
+- H_5: HKDF_Expand-SHA256
```
chris-wood: It's typically written as `HKDF-Expand`, not `HKDF_Expand`. Can we use that instead?

draft-sullivan-cfrg-oprf.md
```diff
@@ -1120,9 +1125,8 @@ Z_0: 043ab5ccb690d844dcb780b2d9e59126d62bc853ba01b2c339ba1c1b78c03e4b6adc5402f7\
 Z_1: 04647e1ab7946b10c1c1c92dd333e2fc9e93e85fdef5939bf2f376ae859248513e0cd91115\
```
chris-wood: Are these test vectors correct with HKDF? We should probably remove them, temporarily, if not.

draft-sullivan-cfrg-oprf.md
```diff
@@ -923,22 +922,28 @@ Steps:
 ## Modified protocol execution
 
 The VOPRF protocol from Section {{protocol}} changes to allow specifying
-multiple blinded PRF inputs [Mi] for i in 1...n. Then P computes the array [Zi]
-and replaces DLEQ_Generate with Batched_DLEQ_Generate over these arrays. The
+multiple blinded PRF inputs [ Mi ] for i in 1...n. Then P computes the array [Zi
```
chris-wood:
```suggestion
multiple blinded PRF inputs [ Mi ] for i in 1...n. Then P computes the array [Zi]
```

draft-sullivan-cfrg-oprf.md
```diff
@@ -923,22 +922,28 @@ Steps:
...
+] and replaces DLEQ_Generate with Batched_DLEQ_Generate over these arrays. The
```
chris-wood:
```suggestion
and replaces DLEQ_Generate with Batched_DLEQ_Generate over these arrays. The
```

draft-sullivan-cfrg-oprf.md
alxdavids: I don't think either is preferred, they're just examples of functions that you might want to use. We give ciphersuites below that use HKDF_Expand, is that enough?

draft-sullivan-cfrg-oprf.md
alxdavids: Good point, I will add this.

draft-sullivan-cfrg-oprf.md
alxdavids: Sure

draft-sullivan-cfrg-oprf.md
alxdavids: Yeah they are correct with HKDF (rather than Shake)

chris-wood:
Thanks for the changes! Once the conflicts are resolved, feel free to merge.

PR #71, opened 2020-03-04
Remove client/server integration advice.

chris-wood:
Upon re-review, this seems to distract from the core of this document. Maybe we can move this the PrivacyPass document? (I note that neither hash-to-curve nor HPKE have such integration advice. They just provide the core primitives.) 

@grittygrease, what do you think?


alxdavids:
I'm happy with removing this

PR #78, opened 2020-05-28
Some clarifications, but still needs more work to be done.

armfazh:
Some comments for improving draft.
Happy to hear your thoughts.

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -570,15 +570,16 @@ The security of our construction is discussed in more detail in
 This protocol may be decomposed into a series of steps, as described
 below:
 
-- Setup(L): Let GG be a group of prime-order p such that the discrete logarithm
-  problem in GG can be solved in 2^L operations. The Prover randomly samples a
-  secret scalar k in GF(p).
-- Blind(x): This operation allows the Verifier to compute M, a blinded
-  representation of x in GG. To do so, the Verifier requires a blind r, which is a scalar uniformly sampled from GF(p).
-- Evaluate(k,M): The Prover uses the secret key k to produce a commitment Z
-  from input M.
-- Unblind(r,Z): The verifier unblinds Z, the blinded OPRF evaluation of the
-  Prover, with blind r yielding N, and outputs N.
+- Setup(l): Let GG=GG(l) be a group with a prime-order p=p(l) (e.g., p
+  is l-bits long). Randomly sample an integer k in GF(p) and output
+  (k,GG)
+- Blind(x): Compute and return a blind, r, and blinded representation of
+  x in GG, denoted M.
+- Evaluate(k,M,h?): Evaluates on input M using secret key k to produce
+  Z, the input h is optional and equal to the cofactor of an elliptic
+  curve. If h is not provided then it defaults to 1.
+- Unblind(r,Z): Unblind blinded OPRF evaluation Z with blind r, yielding
+  N and output N.
 - Finalize(x,N,aux?): Finalize N by first computing dk := H_2(DST, x ..
   N). Subsequently output y := H_2(dk, aux), where aux is some auxiliary
   data encoded as a byte string. If aux is not specified, it defaults to
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -519,11 +519,13 @@ parameter. Let k be the prover's secret key, and Y = kG be its
 corresponding 'public key' for some fixed generator G taken from the
 description of the group GG. This public key Y is also referred to as a
 commitment to the OPRF key k, and the pair (G,Y) as a commitment pair.
-Let x be an arbitrary-length binary string representing the verifier's input to the OPRF protocol.
+Let x be an arbitrary-length byte string representing the verifier's input to the OPRF protocol.
 
 The OPRF protocol begins with V blinding its input for the OPRF
 evaluator such that it appears uniformly distributed from GG. The verifier then
-multiplies the blinded value by its secret key and returns the resulting element. To finish the procotol, V then removes its blind and uses H_2 to hash the result (along with a domain separating label DST) yielding an output.
+multiplies the blinded value by its secret key and returns the resulting element. 
+To finish the protocol, V then removes its blind and uses H_2 to hash the result (along 
+with a domain separating label DST) yielding an output.
 The protocol is illustrated below.
 
 ~~~
@@ -544,7 +546,7 @@ The protocol is illustrated below.
 ~~~
 
 Steps that are enclosed in square brackets (DLEQ_Generate and
-DLEQ_Verify) are REQUIRED for achieving verifiability. These steps are
+DLEQ_Verify) are REQUIRED for verifiability. These steps are
 described in {{dleq}}. In the verifiable mode, we assume that P has
 previously committed to their choice of key k with some values (G,Y=kG)
 and these are publicly known by V. Notice that revealing (G,Y) does not
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -272,9 +272,9 @@ cooperate in computing F(k, x) with P's secret key k and V's input x
 such that: V learns F(k, x) without learning anything about k; and P
 does not learn anything about x. A Verifiable OPRF (VOPRF) is an OPRF
 wherein P can prove to V that F(k, x) was computed using key k, which is
-bound to a trusted public key Y = kG. Informally, this is done by
+bound to a trusted public key Y = kG, and G is the generator of a group. Informally, this is done by
 presenting a non-interactive zero-knowledge (NIZK) proof of equality
-between (G, Y) and (Z, M), where Z = kM for some point M.
+between (G, Y) and (Z, M), where Z and M are group elements such that Z = kM.
 
 OPRFs have been shown to be useful for constructing: password-protected
 secret sharing schemes {{JKK14}}; privacy-preserving password stores
@@ -290,7 +290,7 @@ also been used for password-protected secret sharing schemes e.g.
 
 This document introduces an OPRF protocol built in prime-order groups,
 applying to finite fields of prime-order and also elliptic curve (EC)
-settings. The protocol has the option of being extended to a VOPRF with
+groups. The protocol has the option of being extended to a VOPRF with
 the addition of a NIZK proof for proving discrete log equality
 relations. This proof demonstrates correctness of the computation using
 a known public key that serves as a commitment to the server's secret
@@ -348,7 +348,7 @@ document is structured as follows:
 The following terms are used throughout this document.
 
 - PRF: Pseudorandom Function.
-- OPRF: Oblivious PRF.
+- OPRF: Oblivious Pseudorandom Function.
 - VOPRF: Verifiable Oblivious Pseudorandom Function.
 - Verifier (V): Protocol initiator when computing F(k, x), also known as
   client.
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-voprf.md
```diff
@@ -504,74 +504,59 @@ to the {{JKK14}} construction. With batched proofs ({{batch}}) our
...
-In this section we describe the OPRF and VOPRF protocols. Recall that
-such a protocol takes place between a verifier (V) and a prover (P). We
-may commonly think of the verifier as the client, and the prover as the
-server in the interaction (we will use these names interchangeably
-throughout). The server holds a key k for a PRF. The protocol allows the
-client to learn PRF evaluations on chosen inputs x without revealing x
-to the server.
+## Design
 
-Our OPRF construction is based on the VOPRF construction known as
```
alxdavids, line 515: I think this paragraph can be reworded, but I think it may retain importance for situating the security guarantees provided by the protocol. What do you think @chris-wood?

draft-irtf-cfrg-voprf.md
```diff
@@ -504,74 +504,59 @@ to the {{JKK14}} construction. With batched proofs ({{batch}}) our
...
+                                  Z = kM
                                   [D = DLEQ_Generate(k,G,Y,M,Z)]
                           Z[,D]
                         <-------
     [b = DLEQ_Verify(G,Y,M,Z,D)]
-    N = Zr^(-1) mod p
-    Output H_2(DST, x .. N) mod p [if b=1, else "error"]
+    N = r^(-1)Z
+    h = H_2(DST, x .. N) mod p
+    Output h [if b=false, abort]
```
alxdavids, line 543:
```suggestion
    Output h [if b=false: abort]
```

draft-irtf-cfrg-voprf.md
```diff
@@ -585,20 +570,19 @@ The security of our construction is discussed in more detail in
...
-- Evaluate(k,M,h?): Evaluates on input M using secret key k to produce
-  Z, the input h is optional and equal to the cofactor of an elliptic
-  curve. If h is not provided then it defaults to 1.
-- Unblind(r,Z): Unblind blinded OPRF evaluation Z with blind r, yielding
-  N and output N.
+- Setup(L): Let GG be a group of prime-order p such that the discrete logarithm
+  problem in GG can be solved in 2^L operations. The Prover randomly samples a
+  secret scalar k in GF(p).
+- Blind(x): This operation allows the Verifier to compute M, a blinded
+  representation of x in GG. To do so, the Verifier requires a blind r, which is a scalar uniformly sampled from GF(p).
```
alxdavids, line 577:
```suggestion
  representation of x in GG. To do so, the Verifier requires sampling a blind r, which is a scalar value that must be uniformly sampled from GF(p).
```

draft-irtf-cfrg-voprf.md
```diff
@@ -719,41 +703,30 @@ by evaluating H_2 over dk and auxiliary data aux.
...
-case, and so we focus only on the cases where GF(p) is indeed the base
-field.
-
-Unless otherwise stated, we will always assume that the generator G that
-we use for the group GG is a fixed generator. This generator should be
-available to both the client and the server ahead of the protocol, or
-derived for each different group instantiation using a fixed method. In
-the elliptic curve setting, we recommend using the fixed generators that
-are given as part of the curve description.
+A secure instance is to choose GG as the group of squares of a finite field FF.
```
alxdavids, line 712:
```suggestion
A secure instance includes choosing GG as the group of squares of a finite field FF.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -719,41 +703,30 @@ by evaluating H_2 over dk and auxiliary data aux.
...
+Let FF be field of order q = 2p+1 such that p and q are primes, then the
+squares of FF (elements u^2 where u is an element of FF) form a cyclic group of
+order p.
+
+A more efficient and secure way to instance the (V)OPRF protocol is using
```
alxdavids, line 717:
```suggestion
A more efficient and secure way to instantiate the (V)OPRF protocol is using
```

alxdavids: What does "more secure" mean here?

draft-irtf-cfrg-voprf.md
```diff
@@ -585,20 +570,19 @@ The security of our construction is discussed in more detail in
...
-  (k,GG)
-- Blind(x): Compute and return a blind, r, and blinded representation of
-  x in GG, denoted M.
-- Evaluate(k,M,h?): Evaluates on input M using secret key k to produce
-  Z, the input h is optional and equal to the cofactor of an elliptic
-  curve. If h is not provided then it defaults to 1.
-- Unblind(r,Z): Unblind blinded OPRF evaluation Z with blind r, yielding
-  N and output N.
+- Setup(L): Let GG be a group of prime-order p such that the discrete logarithm
+  problem in GG can be solved in 2^L operations. The Prover randomly samples a
```
alxdavids, line 574: This isn't strictly true, we require hardness from a different assumption. I think it's best to remove the bit about discrete logarithms for now, and I'll completely rework how the Setup phase occurs.

draft-irtf-cfrg-voprf.md
chris-wood: Indeed, and it should probably go to the security considerations. Putting it up front can be a bit cumbersome!

draft-irtf-cfrg-voprf.md
chris-wood: I don't think we should take this change, as it seems to prohibit fixed-base blinding. Let's revert to the old definition, which permitted additive and multiplicative blindings?

draft-irtf-cfrg-voprf.md
chris-wood, line 712: Can we do away with this section entirely and just require that groups have prime order, a fixed generator, and can be implemented efficiently? We can then list examples of such groups, ideally in a table with relevant parameters?

draft-irtf-cfrg-voprf.md
armfazh: I set to be removed as is almost duplicated in Line 499.

draft-irtf-cfrg-voprf.md
armfazh: Agree, it does not sound good.

armfazh, line 717:
```suggestion
A more efficient way to instanciate the (V)OPRF protocol is using
```

draft-irtf-cfrg-voprf.md
armfazh: reverted

draft-irtf-cfrg-voprf.md
armfazh: reverted

draft-irtf-cfrg-voprf.md
```diff
@@ -504,74 +504,59 @@ to the {{JKK14}} construction. With batched proofs ({{batch}}) our
...
-construction (including the NIZK DLEQ proofs from {{dleq}}) is identical
-to the {{JKK14}} construction. With batched proofs ({{batch}}) our
-construction differs slightly in that we can perform multiple VOPRF
-evaluations in one go, whilst only constructing one NIZK proof object.
+Define two hash functions H_1 and H_2 as
 
-## Design
+~~~
+H_1: {0,1}^* -> GG
+H_2: {0,1}^u x {0,1}^v -> {0,1}^w
```
chris-wood, line 513: Unrelated to this change, but what is `u` here? Is DST guaranteed to be of size `u`? (2HashDH doesn't specify the hash this way. Maybe we ought to go back to what's in the original paper?)

draft-irtf-cfrg-voprf.md
```diff
@@ -504,74 +504,59 @@ to the {{JKK14}} construction. With batched proofs ({{batch}}) our
...
-the protocol are modeled as random oracles. Let L be the security
+where H_1 maps arbitrary-length binary strings onto GG, and H_2 maps two
+fixed-length input strings to a fixed-length output string. All hash functions in the protocol are modeled as random oracles. Let L be the security
 parameter. Let k be the prover's secret key, and Y = kG be its
 corresponding 'public key' for some fixed generator G taken from the
 description of the group GG. This public key Y is also referred to as a
 commitment to the OPRF key k, and the pair (G,Y) as a commitment pair.
-Let x be the binary string that is the verifier's input to the OPRF
-protocol (this can be of arbitrary length).
+Let x be an arbitrary-length binary string representing the verifier's input to the OPRF protocol.
```
chris-wood, line 522:
```suggestion
Let x be an arbitrary-length byte string representing the verifier's input to the OPRF protocol.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -504,74 +504,59 @@ to the {{JKK14}} construction. With batched proofs ({{batch}}) our
...
 
 The OPRF protocol begins with V blinding its input for the OPRF
-evaluator such that it appears uniformly distributed GG. The latter then
-applies its secret key to the blinded value and returns the result. To
-finish the computation, V then removes its blind and hashes the result
-(along with a domain separating label DST) using H_2 to yield an output.
-This flow is illustrated below.
+evaluator such that it appears uniformly distributed from GG. The verifier then
+multiplies the blinded value by its secret key and returns the resulting element. To finish the procotol, V then removes its blind and uses H_2 to hash the result (along with a domain separating label DST) yielding an output.
```
chris-wood, line 526:
```suggestion
multiplies the blinded value by its secret key and returns the resulting element. 
To finish the protocol, V then removes its blind and uses H_2 to hash the result (along 
with a domain separating label DST) yielding an output.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -504,74 +504,59 @@ to the {{JKK14}} construction. With batched proofs ({{batch}}) our
...
-    N = Zr^(-1) mod p
-    Output H_2(DST, x .. N) mod p [if b=1, else "error"]
+    N = r^(-1)Z
+    h = H_2(DST, x .. N) mod p
+    Output h [if b=false: abort]
 ~~~
 
 Steps that are enclosed in square brackets (DLEQ_Generate and
-DLEQ_Verify) are optional for achieving verifiability. These are
+DLEQ_Verify) are REQUIRED for achieving verifiability. These steps are
```
chris-wood, line 547:
```suggestion
DLEQ_Verify) are REQUIRED for verifiability. These steps are
```

draft-irtf-cfrg-voprf.md
```diff
@@ -598,7 +583,7 @@ below:
 - Finalize(x,N,aux?): Finalize N by first computing dk := H_2(DST, x ..
   N). Subsequently output y := H_2(dk, aux), where aux is some auxiliary
```
chris-wood, line 584: Unrelated to this change, but this doesn't match the diagram, where we output `H_2(DST, ...)` rather than `H_2(dk, ...)`.

draft-irtf-cfrg-voprf.md
```diff
@@ -719,41 +704,30 @@ by evaluating H_2 over dk and auxiliary data aux.
 The Finalize procedure accepts optional auxiliary byte string input (aux) as
 a means of modifying the PRF output. This parameter SHOULD be used for domain
 separation in (V)OPRF the protocol. Specifically, any system which has
-multiple (V)OPRF applications should use separate aux values to to ensure
-finalized outputs are separate. Guidance for constructing aux can be found
-in {{I-D.irtf-cfrg-hash-to-curve}}; Section 3.1.
+multiple (V)OPRF applications should use different aux values to ensure
+finalized outputs are separate. Guidance for constructing domain separation
+string aux can be found in {{I-D.irtf-cfrg-hash-to-curve}}; Section 3.1.
```
chris-wood, line 709: Should we unify around DST instead of `aux`?

draft-irtf-cfrg-voprf.md
```diff
@@ -719,41 +704,30 @@ by evaluating H_2 over dk and auxiliary data aux.
...
+group can be used as GG if its order is prime. Note, however, that some
+elliptic curves generate groups of composite-order. In this case, GG is usually
+chosen as the largest prime-order subgroup of the elliptic curve. Alternatively,
+GG can be instantiated as a quotient of groups, such as Ristretto {{RISTRETTO}}
+and Decaf {{DECAF}} groups.
+
+Unless otherwise stated, it is assumed that the generator G of the group GG is
+a fixed generator. This generator should be available to both the client and
+the server ahead of the protocol, or derived for each different group
+instantiation using a fixed method. In most of cases, the fixed generators are given as part of the group description.
```
chris-wood, line 730: Should we include a table that lists the group (name) and its parameters, e.g., the generator, base field prime order, etc.?

draft-irtf-cfrg-voprf.md
armfazh: Tracked on #89

draft-irtf-cfrg-voprf.md
armfazh: Tracked in #83

draft-irtf-cfrg-voprf.md
armfazh: Tracked in #90

draft-irtf-cfrg-voprf.md
armfazh: Tracked in #89

armfazh:
Most suggestions were added and some other comments are now tracked as Issues.

PR #98, opened 2020-06-15
Add 256-bit curves back (for OPAQUE).

chris-wood:
Static DH oracles are not a concern for OPAQUE, so, let's add back these curves!

cc @armfazh

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -442,7 +442,7 @@ Specifically, for any elements `A` and `B` that are members of the group
 `GG`, `A + B = B + A` is also a member of `GG`. Scalar multiplication
 (*) is an efficient method for repeated addition operations. Given a
 scalar `r` in `GF(p)` and element `A` in `GG`, `r*A = A + ... + A` (`r`
-times). 
+times).
 
 Note that prime-order groups also define an inverse function such that
 the following property holds:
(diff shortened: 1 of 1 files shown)


alxdavids:
In principle I think this could be okay, but there are some things that I'd like to have discussion on.

1. Do we want to explicitly change the curve25519 and curve448 ciphersuites to actually reference ristretto and decaf? Effectively what we are currently saying is that you need to create a prime-order group and you can do this using ristretto and decaf, but maybe we need to be more explicit in only supporting these instantiations? The reason being that we may end up with a load of un-interoperable implementations with implementers that interpret the spec differently.

2. I wonder whether we should be explicit on how many bits of security we are offering with each ciphersuite in the case of a static-DH oracle attack? Because P256 and curve25519 are affected differently, and so a curve25519 implementation would actually retain more security, but this is not obvious when just looking at the two group instantiations in isolation.

chris-wood:
> Do we want to explicitly change the curve25519 and curve448 ciphersuites to actually reference ristretto and decaf? Effectively what we are currently saying is that you need to create a prime-order group and you can do this using ristretto and decaf, but maybe we need to be more explicit in only supporting these instantiations? The reason being that we may end up with a load of un-interoperable implementations with implementers that interpret the spec differently.

There's no specification for Decaf, yet, so I'm not sure. Until that's true, I'd prefer we have an entirely separate ciphersuite for Ristretto, which uses the FROM_UNIFORM_BYTES function accordingly. 

> I wonder whether we should be explicit on how many bits of security we are offering with each ciphersuite in the case of a static-DH oracle attack? Because P256 and curve25519 are affected differently, and so a curve25519 implementation would actually retain more security, but this is not obvious when just looking at the two group instantiations in isolation.

I don't think this is necessary. We already discuss it in the security considerations.

PR #99, opened 2020-06-16
Add initial Sage reference implementation

chris-wood:
cc @armfazh


alxdavids:
LGTM 👍 I think we can tackle verifiability in future PRs.

PR #102, opened 2020-06-17
Add suffix-free domain separation for H_1.

chris-wood:
See https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-08#section-10.3.

### DIFF
--- a/poc/oprf.sage
+++ b/poc/oprf.sage
@@ -164,12 +164,14 @@ class Client(object):
 		finalize_dst = _as_bytes("RFCXXXX-Finalize")
 		encoded_point = self.suite.group.serialize(y)
 
-		h.update(I2OSP(len(finalize_dst), 2))
-		h.update(finalize_dst)
 		h.update(I2OSP(len(x), 2))
 		h.update(x)
 		h.update(I2OSP(len(encoded_point), 2))
 		h.update(encoded_point)
+		h.update(I2OSP(len(aux), 2))
+		h.update(aux)
+		h.update(I2OSP(len(finalize_dst), 2))
+		h.update(finalize_dst)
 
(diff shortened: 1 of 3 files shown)


poc/oprf.sage
```diff
@@ -164,12 +164,14 @@ class Client(object):
...
 		encoded_point = self.suite.group.serialize(y)
 
-		h.update(I2OSP(len(finalize_dst), 2))
-		h.update(finalize_dst)
 		h.update(I2OSP(len(x), 2))
 		h.update(x)
 		h.update(I2OSP(len(encoded_point), 2))
 		h.update(encoded_point)
+		h.update(I2OSP(len(aux), 2))
+		h.update(aux)
```
alxdavids, line 172: shall we rename this to aux as well?

chris-wood: Hmm? It is already named aux.

armfazh:
Does the purpose of the `info` variable must be user-specific, rather than use it as domain separation tag? 
One can define a separation tag without requiring the user to care about it. do you think this could work?

poc/oprf.sage
alxdavids: oops, I meant `info`.

chris-wood: Doh, yes, good catch! Fixed.

chris-wood:
> Does the purpose of the info variable must be user-specific, rather than use it as domain separation tag?

`info` is an application-specific domain separation string. If absent, we still require domain separation, so there's "two" layers of separation in play here.

chris-wood:
@alxdavids if you're OK with these changes, please feel free to merge!

PR #104, opened 2020-06-19
Fixes #101: use compressed point encoding

### LINKED ISSUE #101: Compressed or uncompressed point encoding?
Should we support both, or just one?

alxdavids:


poc/oprf.sage
```diff
@@ -31,212 +32,222 @@ def to_hex(octet_string):
...
+    def random_scalar(self):
+        return self.F.random_element()
+
+    def identity(self):
+        return self.curve.random_element() * self.order()
+
+    def serialize(self, element):
+        x, y = element[0], element[1]
+        sgn = sgn0(y)
+        byte = (2, 3)[sgn == 0]
```
chris-wood, line 95: It took me a second to figure out what was happening here -- could we simplify?

```
b = 3 if sgn == 0 else 2
```

alxdavids: Sure I can simplify

alxdavids:
Rebased and simplified so will merge.

PR #107, opened 2020-06-19
Fixes #79

### LINKED ISSUE #79: Specify domain-separated hashes
By following the recommendations given in the hash to curve draft, this draft must specify how to domain-separate all the hash functions required.

alxdavids:
- Removes H3 as it is the same as H1
- Add prefix-free domain separation for hash function calls

cc @armfazh

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -955,16 +955,13 @@ descriptions. Note that both algorithms create a batched proof for
 multiple evaluations of the VOPRF.
 
 In all three algorithms that we detail below we make the following
-domain separation labels available as global variables:
+domain separation labels available as global variables.
 
 ~~~
-opaque challengeDST<1..2^16-1>
-opaque seedDST<1..2^16-1>
-opaque compositeDST<1..2^16-1>
+opaque challengeDST<1..2^16-1> = "RFCXXXX-challenge"
+opaque seedDST<1..2^16-1> = "RFCXXXX-seed"
+opaque compositeDST<1..2^16-1> = "RFCXXXX-composite"
 ~~~
-
-Each of these variables MUST be chosen by the Server implementation.
-In addition, the choice SHOULD be made available to the Client.
 
 ## GenerateProof
 
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -954,13 +954,14 @@ The proof generation and verification algorithms are denoted by
 descriptions. Note that both algorithms create a batched proof for
 multiple evaluations of the VOPRF.
 
-In all three algorithms that we detail below we make the following
-domain separation labels available as global variables.
+In the algorithms below we set the following domain separation labels.
+These are used to control the domains of the hash function evaluations
+that we use.
 
 ~~~
-opaque challengeDST<1..2^16-1> = "RFCXXXX-challenge"
-opaque seedDST<1..2^16-1> = "RFCXXXX-seed"
-opaque compositeDST<1..2^16-1> = "RFCXXXX-composite"
+opaque challengeDST<1..2^16-1>
+opaque seedDST<1..2^16-1>
+opaque compositeDST<1..2^16-1>
 ~~~
 
 ## GenerateProof
@@ -986,13 +987,14 @@ Steps:
  5.  if (r == 0): go back to the previous step
  6.  a3 = GG.Serialize(r * G)
  7.  a4 = GG.Serialize(rM)
- 8.  h2Input = I2OSP(len(gen), 2) || gen || I2OSP(len(pkS), 2) || pkS ||
+ 8.  challengeDST = "RFCXXXX-challenge"
+ 9.  h2Input = I2OSP(len(gen), 2) || gen || I2OSP(len(pkS), 2) || pkS ||
             I2OSP(len(a1), 2) || a1 || I2OSP(len(a2), 2) || a2 ||
             I2OSP(len(a3), 2) || a3 || I2OSP(len(a4), 2) || a4 ||
             I2OSP(len(challengeDST), 2) || challengeDST
- 9.  c = H_2(h2Input) mod p
- 10. s = (r - c * skS) mod p
- 11. Output (c, s)
+ 10. c = H_2(h2Input) mod p
+ 11. s = (r - c * skS) mod p
+ 12. Output (c, s)
 ~~~
 
 We note here that it is essential that a different r value is used for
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-voprf.md
```diff
@@ -769,7 +767,8 @@ Output:
 Steps:
 
  1. DST = "RFCXXXX-Finalize"
```
chris-wood, line 769:
```suggestion
 1. finalizeDST = "RFCXXXX-Finalize"
```

draft-irtf-cfrg-voprf.md
```diff
@@ -931,8 +930,19 @@ procedures: GenerateProof and VerifyProof. These are specified below.
...
+In all three algorithms that we detail below we make the following
+domain separation labels available as global variables:
+
+~~~
+opaque challengeDST<1..2^16-1>
+opaque seedDST<1..2^16-1>
+opaque compositeDST<1..2^16-1>
+~~~
+
+Each of these variables should be chosen by the Server implementation.
```
chris-wood, line 944:
```suggestion
Each of these variables MUST be chosen by the Server implementation.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -931,8 +930,19 @@ procedures: GenerateProof and VerifyProof. These are specified below.
...
+In addition, the choice should be made available to the Client.
```
chris-wood, line 945:
```suggestion
In addition, the choice SHOULD be made available to the Client.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -952,16 +962,19 @@ Steps:
...
  6.  a3 = GG.Serialize(r * G)
  7.  a4 = GG.Serialize(rM)
- 8.  c = H_2(gen || public_key || a1 || a2 || a3 || a4) mod p
- 9.  s = (r - ck) mod p
- 10. Output (c, s)
+ 8.  h2Input = I2OSP(len(gen), 2) || gen || I2OSP(len(public_key), 2) ||
+            public_key || I2OSP(len(a1), 2) || a1 ||
+            I2OSP(len(a2), 2) || a2 || I2OSP(len(a3), 2) || a3 ||
+            I2OSP(len(a4), 2) || a4 || I2OSP(len(challengeDST), 2) ||
+            challengeDST
```
chris-wood, line 974:
```suggestion
 8.  h2Input = I2OSP(len(gen), 2) || gen || 
               I2OSP(len(public_key), 2) || public_key || 
               I2OSP(len(a1), 2) || a1 || I2OSP(len(a2), 2) || a2 || 
               I2OSP(len(a3), 2) || a3 || I2OSP(len(a4), 2) || a4 || 
               I2OSP(len(challengeDST), 2) || challengeDST
```

draft-irtf-cfrg-voprf.md
```diff
@@ -990,15 +1003,18 @@ Steps:
...
  5. B' = (proof[1] * M + proof[0] * Z)
  6. a3 = GG.Serialize(A')
  7. a4 = GG.Serialize(B')
- 8. c  = H_2(gen || public_key || a1 || a2 || a3 || a4) mod p
- 9. Output c == proof[0] mod p
+ 8.  h2Input = I2OSP(len(gen), 2) || gen || I2OSP(len(public_key), 2) ||
+            public_key || I2OSP(len(a1), 2) || a1 ||
+            I2OSP(len(a2), 2) || a2 || I2OSP(len(a3), 2) || a3 ||
+            I2OSP(len(a4), 2) || a4 || I2OSP(len(challengeDST), 2) ||
+            challengeDST
```
chris-wood, line 1015:
```suggestion
 8.  h2Input = I2OSP(len(gen), 2) || gen || 
            I2OSP(len(public_key), 2) || public_key || 
            I2OSP(len(a1), 2) || a1 || I2OSP(len(a2), 2) || a2 || 
            I2OSP(len(a3), 2) || a3 || I2OSP(len(a4), 2) || a4 || 
            I2OSP(len(challengeDST), 2) || challengeDST
```

draft-irtf-cfrg-voprf.md
chris-wood: (Noting for later -- `h2Input` case is inconsistent with `public_key` case.)

draft-irtf-cfrg-voprf.md
```diff
@@ -1013,31 +1029,36 @@ Input:
...
+ 4. M = GG.Identity()
+ 5. Z = GG.Identity()
+ 6. for i = 0 to m:
     1. di = 1
     2. Mi = GG.Deserialize(blinded_tokens[i])
     3. Zi = GG.Deserialize(ev.elements[i])
     4. if (m > 1):
-       1. di = H_2(seed || i' || DST_dleq)
-       2. if (di > GG.order()):
+       1. h2Input = I2OSP(len(seed), 2) || seed || I2OSP(i', 2) ||
```
chris-wood, line 1052: Do we need to encode the length of `seed` here, as it's fixed-length based on H_1? (I don't see the harm in doing so. Just thinking out loud.)

draft-irtf-cfrg-voprf.md
```diff
@@ -931,8 +930,19 @@ procedures: GenerateProof and VerifyProof. These are specified below.
 The proof generation and verification algorithms are denoted by
 `GenerateProof` and `VerifyProof` respectively, see below for
 descriptions. Note that both algorithms create a batched proof for
-multiple evaluations of the VOPRF. Note further that both algorithms can
-be domain-separated using the global `opaque DST_dleq<1..2^16-1>` value.
+multiple evaluations of the VOPRF.
+
+In all three algorithms that we detail below we make the following
```
chris-wood, line 935: Since these are domain separators for the protocol, shouldn't we define them here, rather than punting to the application? We do this for `Finalize()` above, for example.

alxdavids:
> Nice reduction :) I only have one blocking comment on the DSTs. Namely, who specifies them? I think it should be this specification, and here's why: if an application chooses to not set these to anything, i.e., they're some default string or they're the same string across challenge, seed, and composite, then they don't do their job of effectively cloning the oracles. If the protocol specifies them as "RFCXXXX-Challenge", "RFXXXX-Seed", and "RFCXXX-Composite", for example, the application gets this separation for free.

Thanks for the review! Yeah you make a good point. I agree with this change, I'll make it and update the commit.

draft-irtf-cfrg-voprf.md
alxdavids: Since we specify `H_2` as part of the ciphersuite, I think it's worth being consistent in case we decide to change it.

draft-irtf-cfrg-voprf.md
alxdavids: Yep, will address this

draft-irtf-cfrg-voprf.md
alxdavids: Made this change

draft-irtf-cfrg-voprf.md
```diff
@@ -938,8 +937,16 @@ procedures: GenerateProof and VerifyProof. These are specified below.
...
 descriptions. Note that both algorithms create a batched proof for
-multiple evaluations of the VOPRF. Note further that both algorithms can
-be domain-separated using the global `opaque dleqDST<1..2^16-1>` value.
+multiple evaluations of the VOPRF.
+
+In all three algorithms that we detail below we make the following
+domain separation labels available as global variables.
+
+~~~
+opaque challengeDST<1..2^16-1> = "RFCXXXX-challenge"
```
chris-wood, line 946: Should we move these to the functions where they're used, as we did with Finalize?

alxdavids: Sure, just made that change now

PR #161, opened 2020-10-16
Change placeholder DST.

### LINKED ISSUE #143: Use a draft-specific placeholder DST string
If folks deploy experimental systems using VOPRF drafts, the fact that the DST uses "RFCXXXX" might break domain separation between drafts. HPKE used to use the same placeholder but switched to "HPKE-05", etc., with an RFC editor note:

> [[RFC editor: please change "HPKE-05" to "RFCXXXX", where XXXX is the final number, before publication.]]

See https://github.com/cfrg/draft-irtf-cfrg-hpke/pull/138.

Perhaps VOPRF should do something similar here.

chris-wood:
Closes #143.

cc @davidben

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -293,7 +293,7 @@ groups, including elliptic curves.
 
 A pseudorandom function (PRF) F(k, x) is an efficiently computable
 function taking a private key k and a value x as input. This function is
-pseudorandom if the keyed function K(_) = F(K, _) is indistinguishable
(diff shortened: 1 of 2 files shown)


alxdavids (approved):
Looks good! Do we need something like this somewhere:

> 

    [[RFC editor: please change "HPKE-05" to "RFCXXXX", where XXXX is the final number, before publication.]]

as mentioned in #143?

chris-wood:
That's already in the change:

```
[[RFC editor: please change "VOPRF05" to "RFCXXXX", where XXXX is the final number, here and elsewhere before publication.]]
```

PR #166, opened 2020-10-19
Fix a seed used to derive all test vectors.

chris-wood:
cc @claucece


claucece:
Thank you! Looks great.

PR #172, opened 2020-10-28
Clean up ciphersuite section. Fix some references too.

chris-wood:
This does a couple things:

- Removes `Order()` -- it was never used.
- Unifies Serialization content. 
- Removes scalar multiplication instructions, as I think these are unnecessary.

It also fixes some references (uses updated versions and removes unused references).

cc @claucece


claucece:
LGTM. Thanks!

PR #173, opened 2020-10-28
replaces client by self in contextString

armfazh:


chris-wood (approved):
Good catch -- I didn't notice this.

PR #176, opened 2020-10-30
Refactor to remove special-case ristretto255/decaf448 paths.

chris-wood:
cc @claucece, @armfazh

### DIFF
--- a/poc/oprf.sage
+++ b/poc/oprf.sage
@@ -171,8 +171,7 @@ class GroupRistretto255(Group):
         return Ed25519Point().decode(encoded)
 
     def hash_to_group(self, msg, dst):
-        uniform_bytes = expand_message_xmd(msg, dst, 64, hashlib.sha256, self.k)
-        return Ed25519Point().map(uniform_bytes)
+        return Ed25519Point().hash_to_group(msg, dst)
 
     def hash_to_scalar(self, msg, dst=""):
         return hash_to_field(msg, 1, dst, self.order(), 1, self.L, expand_message_xmd, hashlib.sha256, self.k)[0][0]
@@ -199,8 +198,7 @@ class GroupDecaf448(Group):
         return Ed448GoldilocksPoint().decode(encoded)
 
     def hash_to_group(self, msg, dst):
-        uniform_bytes = expand_message_xmd(msg, dst, 112, hashlib.sha512, self.k)
-        return Ed448GoldilocksPoint().map(uniform_bytes)
+        return Ed448GoldilocksPoint().hash_to_group(msg, dst)
 
     def hash_to_scalar(self, msg, dst=""):
         return hash_to_field(msg, 1, dst, self.order(), 1, self.L, expand_message_xmd, hashlib.sha512, self.k)[0][0]
--- a/poc/oprf.sage
+++ b/poc/oprf.sage
@@ -34,10 +34,16 @@ class Group(object):
     def __init__(self, name):
         self.name = name
 
+    def generator(self):
+        return None
+
     def random_scalar(self):
         return None
 
     def identity(self):
+        return 0
+
+    def order(self):
         return 0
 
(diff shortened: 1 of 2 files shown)


armfazh (approved):

poc/oprf.sage
```diff
@@ -137,6 +149,62 @@ class GroupP521(GroupNISTCurve):
...
+    def identity(self):
+        return Ed448GoldilocksPoint().identity()
+
+    def serialize(self, element):
+        return element.encode()
+
+    def deserialize(self, encoded):
+        return Ed448GoldilocksPoint().decode(encoded)
+
+    def hash_to_group(self, msg, dst):
```
claucece, line 201: Per ristretto et al draft, it should be:

```
        u = expand_message_xof(msg, dst, int(112), hashlib.shake_256, 224)
        P1 = self.map(u[0:32])
        P2 = self.map(u[32:64])
        P = P1 + P2
```

poc/oprf.sage
```diff
@@ -137,6 +149,62 @@ class GroupP521(GroupNISTCurve):
...
+        uniform_bytes = expand_message_xmd(msg, dst, 112, hashlib.sha512, self.k)
```
claucece, line 202: I'm not sure the internals of this.. but it need to be expanded to a 112 array, which will no be given with sha512. That is why it was using shake256.

poc/oprf.sage
```diff
@@ -137,6 +149,62 @@ class GroupP521(GroupNISTCurve):
...
+    def identity(self):
+        return Ed25519Point().identity()
+
+    def serialize(self, element):
+        return element.encode()
+
+    def deserialize(self, encoded):
+        return Ed25519Point().decode(encoded)
+
+    def hash_to_group(self, msg, dst):
```
claucece, line 173: Per ristretto et al draft, it should be:

```
        u = expand_message_xmd(msg, dst, int(64), hashlib.sha512, 128)
        P1 = self.map(u[0:32])
        P2 = self.map(u[32:64])
        P = P1 + P2
        return P
```

poc/oprf.sage
```diff
@@ -137,6 +149,62 @@ class GroupP521(GroupNISTCurve):
...
+        uniform_bytes = expand_message_xmd(msg, dst, 64, hashlib.sha256, self.k)
```
claucece, line 174: Here we need to expand it to 64, which I'm not sure the sha256 will give.

poc/oprf.sage
chris-wood: No, see: https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10#appendix-B

poc/oprf.sage
chris-wood: See above!

poc/oprf.sage
chris-wood: See above!

poc/oprf.sage
claucece: I'm not sure about that. See the comment for the `MAP` function in ristretto et al draft:

```
The MAP function is defined on a 32-bytes string as:
```

The string here is 64. 

Which means that it should use the one-way-map function of the ristretto et al draft, which is:

```
1. Compute P1 as "MAP(b[0:32])".
2. Compute P2 as "MAP(b[32:64])".
3. Return P1 + P2.
```

Meaning that:

```
   1. uniform_bytes = expand_message(msg, DST, 64)
   2. P = ristretto255_map(uniform_bytes)
   3. return P
```

`istretto255_map(uniform_bytes)` refers to the one-way-map. Otherwise, the map will only handle the first 32 bytes.

chris-wood: `ristretto255_map` indeed refers to the one-way map, and not the internal `MAP` function. :)

PR #188, opened 2020-11-05
updates fields in test vectors

armfazh:
I finally get reproducible test vectors, I suspect that the cause was grouping the suites in a set versus in a list.
Merge after #187

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -702,13 +702,11 @@ def GenerateProof(skS, pkS, blindToken, element)
 ##### Batching inputs
 
 Unlike other functions, `ComputeComposites` takes lists of inputs,
-rather than a single input. This allows to produce a combined proof
-in `GenerateProof` for a batch of elements.
-(In the pseudocode above, the single inputs `blindToken`
-and `element` passed as one-item lists to `ComputeComposites`).
-Applications can take advantage of this functionality by invoking
-`GenerateProof` on batches of inputs and producing a `BatchedEvaluation`
-structure as the output of `Evaluate`.
+rather than a single input. Applications can take advantage of this
+functionality by invoking `GenerateProof` on batches of inputs to
+produce a combined, constant-size proof. (In the pseudocode above,
+the single inputs `blindToken` and `element` are passed as one-item
+lists to `ComputeComposites`.)
 
 ##### Fresh randomness
 
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -702,17 +702,13 @@ def GenerateProof(skS, pkS, blindToken, element)
 ##### Batching inputs
 
 Unlike other functions, `ComputeComposites` takes lists of inputs,
-rather than a single input. It is optimized to produce a constant-size
-output. This functionality lets applications batch inputs together to
-produce a constant-size proofs from `GenerateProof`. Applications can
-take advantage of this functionality by invoking `GenerateProof` on
-batches of inputs. (Notice that in the pseudocode above, the single
-inputs `blindToken` and `element` are translated into lists before
-invoking `ComputeComposites`. A batched `GenerateProof` variant would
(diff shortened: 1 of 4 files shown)


claucece:
Oh nice! Thank you! Can you also delete the check for ristretto or decaf on the test file?

draft-irtf-cfrg-voprf.md
```diff
@@ -1195,7 +1203,7 @@ in {{!I-D.irtf-cfrg-hash-to-curve}} when instantiating the function.
 
 ## Timing Leaks
 
-To ensure no information is leaked during protocol execution, all
+To prevent no information is leaked during protocol execution, all
```
chris-wood, line 1206:
```suggestion
To ensure no information is leaked during protocol execution, all
```

draft-irtf-cfrg-voprf.md
```diff
@@ -702,17 +702,13 @@ def GenerateProof(skS, pkS, blindToken, element)
...
-
-Note that using batched inputs creates a `BatchedEvaluation` object as
-the output of `Evaluate`.
+rather than a single input. This allows to produce a combined proof
+in `GenerateProof` for a batch of elements.
+(In the pseudocode above, the single inputs `blindToken`
+and `element` passed as one-item lists to `ComputeComposites`).
+Applications can take advantage of this functionality by invoking
+`GenerateProof` on batches of inputs and producing a `BatchedEvaluation`
+structure as the output of `Evaluate`.
```
chris-wood, line 711:
```suggestion
rather than a single input. Applications can take advantage of this
functionality by invoking `GenerateProof` on batches of inputs to 
produce a combined, constant-size proof. (In the pseudocode above, 
the single inputs `blindToken` and `element` passed as one-item lists 
to `ComputeComposites`.) 
```

draft-irtf-cfrg-voprf.md
```diff
@@ -723,6 +719,11 @@ used.
 
 #### ComputeComposites
 
+The definition of `ComputeComposites` is given below. This function is
```
chris-wood, line 722: Can we refactor this function so that it doesn't have a reference to `self` internally? It's sort of awkward as-is.

poc/test_oprf.sage
```diff
@@ -76,29 +92,25 @@ class Protocol(object):
...
     for suite in oprf_ciphersuites:
-        skS, pkS = KeyGen(suite)
-        server, pkS = SetupVerifiableServer(suite, skS, pkS)
-        client = SetupVerifiableClient(suite, pkS)
-        protocol = Protocol()
-        vectors["Verifiable" + suite.name] = protocol.run(client, server, "test information")
+        for mode in [mode_base, mode_verifiable]:
+            protocol = Protocol(suite, mode)
+            vector = protocol.run()
+            vectors += [ vector ]
```
chris-wood, line 113:
```suggestion
            vectors.append(protocol.run())
```

chris-wood (changes requested):
I quite like the change, but I think it needs to be reformatted for clarity. Also, as @claucece suggests, can we remove the special case ristretto255/decaf448 changes from the test file?

chris-wood:
> I suspect that the cause was grouping the suites in a set versus in a list.

Hah! That does explain it, as I only ran one ciphersuite each time I did the test (to save time).

armfazh:
> Can you also delete the check for ristretto or decaf on the test file?

which check are you referring to?

draft-irtf-cfrg-voprf.md
armfazh: I think this also extends to the contextString?
```
  seedDST = "VOPRF05-seed-" || self.contextString
```

armfazh: I followed the same pattern of using self. Ideas are welcomed.

chris-wood: Ohh, yeah. Hmm. How about we merge as-is and then I make an attempt at clarifying after?

chris-wood (approved):

PR #203, opened 2020-11-10
Move ristretto255 suite to SHA-512 for better alignment with Ed25519.

### LINKED ISSUE #200: Change SHA-256 to SHA-512 for Ristretto255
Hello,

in https://github.com/cfrg/draft-irtf-cfrg-voprf/blob/72ad54bba21da719456d85d3a12304410f83448d/draft-irtf-cfrg-voprf.md#oprfristretto255-sha-256, it is specified to use SHA-256 with Ristretto255. SHA-256 outputs 32 bytes.

https://tools.ietf.org/html/draft-irtf-cfrg-ristretto255-decaf448-00#section-4.3.4 states that the one-way map operates on 64-byte strings.

The hash function to be used should be SHA-512, which outputs 64 bytes, as described in https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10#appendix-B for the `ristretto255_XMD:SHA-512_R255MAP_RO_` cipher suite.

chris-wood:
Closes #200.

cc @bytemare, @jedisct1

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -945,21 +945,21 @@ and serialization functionalities.
 Applications should take caution in using ciphersuites targeting P-256
 and ristretto255. See {{cryptanalysis}} for related discussion.
 
-## OPRF(ristretto255, SHA-256)
+## OPRF(ristretto255, SHA-512)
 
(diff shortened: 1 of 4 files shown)


jedisct1:
Looks good!

armfazh (approved):

bytemare:
Looks good to me :)

PR #210, opened 2020-11-12
Pads bytestrings with zeros and fixes endianess

### LINKED ISSUE #209: Invalid encodings in test vectors
Hello,

I tried testing my implementation in Go with the vectors found here: https://github.com/cfrg/draft-irtf-cfrg-voprf/blob/8ec42c73db97a862827b8eec959a4df3d031adec/poc/vectors/allVectors.json

I encounter errors decoding private key values for Ristretto255. Stripping the 0x, the value is of length 63, which is odd, and Go's `hex.DecodeString()` returns the error `"encoding/hex: odd length hex string"`. I suppose something went wrong with generating these values.

Maybe testing the values generated by https://github.com/cfrg/draft-irtf-cfrg-voprf/blob/8ec42c73db97a862827b8eec959a4df3d031adec/poc/test_oprf.sage would detect this kind of issues earlier.

armfazh:
the `hex` function didn't pad with zeros and issues with endianess.

 - Included a function for padding, so all byte strings are of even length.
 - Removes '0x' prefix because the array of bytes can be either in big-endian (256-384-521) or little-endian order (255-448).
 - batch is an integer.

Fixes #209


bytemare:
Hi,

I just tested with these new padded values, and hex encoding now seems valid. Thanks!

bytemare:
On the other hand, I get an `"invalid scalar encoding"` error when using decoded `skSm` values for Ristretto255 using https://github.com/gtank/ristretto255/blob/9e5f56cbf4b15d3a98c048363d8d52fe11e0e2d0/internal/scalar/scalar.go#L76

poc/groups.sage
```diff
@@ -124,6 +121,9 @@ class GroupNISTCurve(Group):
             y = -y
         return self.curve(self.F(x), self.F(y))
 
+    def serialize_scalar(self, scalar):
+        return I2OSP(scalar%self.order(), int(ceil(len(self.p.bits())/8)))
```
chris-wood, line 125:
```suggestion
        return I2OSP(scalar % self.order(), int(ceil(len(self.p.bits()) / 8)))
```

poc/groups.sage
```diff
@@ -176,6 +173,9 @@ class GroupRistretto255(Group):
     def deserialize(self, encoded):
         return Ed25519Point().decode(encoded)
 
+    def serialize_scalar(self, scalar):
+        return I2OSP(scalar%self.order(), 32)[::-1]
```
chris-wood, line 177:
```suggestion
        return I2OSP(scalar % self.order(), 32)[::-1]
```

poc/groups.sage
```diff
@@ -206,6 +203,9 @@ class GroupDecaf448(Group):
     def deserialize(self, encoded):
         return Ed448GoldilocksPoint().decode(encoded)
 
+    def serialize_scalar(self, scalar):
+        return I2OSP(scalar%self.order(), 56)[::-1]
```
chris-wood, line 207:
```suggestion
        return I2OSP(scalar % self.order(), 56)[::-1]
```

chris-wood:
This seems to work: https://play.golang.org/p/SHILTLDHRZ3

@bytemare, can you please check the vectors?

chris-wood (approved):

PR #217, opened 2020-11-23
Add clarifications about batching.

### LINKED ISSUE #201: Clarifications on BatchedEvaluation
Hello,

reading the latest version of the draft, there are very few explanations about batching inputs/evaluations. For the reader unfamiliar with the subject it is not clear what these are about.

From what I understand from a conversation with @chris-wood:
the mechanism works in a situation where a client requests multiple evaluations and could do it by sending all blinded inputs at once. The server would then evaluate them all but have a single proof for them all, and send this single proof back with the individual evaluations.
So, for N blinded inputs from the client, instead of having N requests, N evaluations, N proofs, and N responses, we would bring that down to  1 proof and 1 request-response carrying the N blinded inputs and N evaluations.

Is this correct?

bytemare:
Closes #201.

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -925,6 +925,21 @@ def Unblind(blind, evaluatedElement, blindedElement, pkS, proof):
   return unblindedElement
 ~~~
 
+### Batching
+
+Clients that need multiple verifiable evaluations would make as much requests
+to the server. Batching the inputs enables the server to evaluate them all
+individually and to compute a single NIZK proof for the whole set at once.
+
+With this technique, the client sends a set of blindedElements to the server,
+and the server responds with the set of evaluatedElements and the proof object
+(proofC and proofS).
+
+The optimization takes place in the ComputeComposites function, and therefore
+benefits both the server and the client. Hence, for N blinded inputs from the
+client, instead of having N requests, N evaluations, N proofs, and N responses,
+we would bring that down to N evaluations and 1 proof in 1 roundtrip.
+
 # Ciphersuites {#ciphersuites}
 
 A ciphersuite (also referred to as 'suite' in this document) for the protocol
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -937,8 +937,9 @@ and the server responds with the set of evaluatedElements and a single proof obj
 
 The optimization takes place in the ComputeComposites function, and therefore
 benefits both the server and the client. Hence, for N blinded inputs from the
-client, instead of having N requests, N evaluations, N proofs, and N responses,
-we would bring that down to N evaluations and 1 proof in 1 roundtrip.
+client, instead of having N roundtrips for fetching N evaluations and N proofs,
+we would bring that down to a single roudtrip carrying N evaluations and 
+one proof.
 
 # Ciphersuites {#ciphersuites}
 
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-voprf.md
```diff
@@ -925,6 +925,21 @@ def Unblind(blind, evaluatedElement, blindedElement, pkS, proof):
   return unblindedElement
 ~~~
 
+### Batching
```
chris-wood, line 928: This text is somewhat redundant with what's in the "Batching inputs" section. Would it be possible to add this paragraph to the end of that section? 

```
In particular, servers can produce a single, constant-sized proof for N 
client inputs, rather than one proof per client input. This optimization 
benefits clients and servers since it amortizes the cost of proof generation 
and bandwidth across multiple requests.
```

chris-wood (changes requested):
Thanks for the text! I think this can be simplified somewhat and merged with already existing text. Please let me know what you think of the suggestion.

draft-irtf-cfrg-voprf.md
```diff
@@ -925,6 +925,21 @@ def Unblind(blind, evaluatedElement, blindedElement, pkS, proof):
...
+individually and to compute a single NIZK proof for the whole set at once.
+
+With this technique, the client sends a set of blindedElements to the server,
+and the server responds with the set of evaluatedElements and the proof object
+(proofC and proofS).
+
+The optimization takes place in the ComputeComposites function, and therefore
+benefits both the server and the client. Hence, for N blinded inputs from the
+client, instead of having N requests, N evaluations, N proofs, and N responses,
+we would bring that down to N evaluations and 1 proof in 1 roundtrip.
```
armfazh, line 941: Feel free to edit as needed.

```suggestion
benefits both the server and the client. Hence, for N blinded inputs from the
client, instead of having N roundtrips for fetching N evaluations and N proofs,
we would bring that down to a single roudtrip carrying N evaluations and 
one proof.
```

draft-irtf-cfrg-voprf.md
armfazh, line 928: Does batching deserves its own section to highlight its relevance?

```suggestion
## Batching
```

armfazh (commented):
Some comments

draft-irtf-cfrg-voprf.md
bytemare: Do you mean dropping this (@armfazh suggestions have been included)

> ## Batching
> 
> Clients that need multiple verifiable evaluations would make as many requests
> to the server. Batching the inputs enables the server to evaluate them all
> individually and to compute a single NIZK proof for the whole set at once.
> 
> With this technique, the client sends a set of blindedElements to the server,
> and the server responds with the set of evaluatedElements and a single proof object
> (proofC and proofS).
> 
> The optimization takes place in the ComputeComposites function, and therefore
> benefits both the server and the client. Hence, for N blinded inputs from the
> client, instead of having N roundtrips for fetching N evaluations and N proofs,
> we would bring that down to a single roundtrip carrying N evaluations and 
> one proof.

and adding your paragraph to `##### Batching inputs` instead? Or merging the suggestions and adding them to `##### Batching inputs`?

chris-wood: I suggest the former! I think the proposed text (with @armfazh's suggestions) have more words than are needed.

bytemare: Indeed, it's much more concise. May I suggest adding the notion of the client sending the batch in a single request, though? So it's undoubtedly clear to the reader it's not about sparse elements the client sends over time.

How about this:
> In particular, servers can produce a single, constant-sized proof for N 
> client inputs sent in a single request, rather than one proof per client
> input. This optimization benefits clients and servers since it amortizes
> the cost of proof generation and bandwidth across multiple requests.

chris-wood: Looks good!

armfazh (approved):

chris-wood (approved):

PR #218, opened 2020-11-24
Add DST to JSON test vectors.

### LINKED ISSUE #215: Add h2c DST in test vectors
I suggest adding the DST used for hash-to-group and hash-to-scalar in the test vectors, just like info. This would allow identifying the correct assembly of that DST before trying to hash anything and fail early.

Since I can't install Sage on my working machine right now, here is the suggested line I think would allow this to be added.

test_oprf.sage, line 121
``` 
vecSuite["skSm"] = to_hex(group.serialize_scalar(server.skS))
+ vecSuite["dst"] = to_hex(_as_bytes("VOPRF06-HashToGroup-") + I2OSP(self.mode, 1) + I2OSP(self.suite.identifier, 2) )
if self.mode == mode_verifiable:
```

chris-wood:
Closes #215.

cc @bytemare


bytemare:
Oh my, I was just doing it this very moment! Thanks :)

poc/ristretto_decaf.sage
```diff
@@ -414,7 +414,8 @@ def testMapDecaf(cls):
     exp = bytearray.fromhex("f4ccb31d263731ab88bed634304956d2603174c66da38742053fa37dd902346c3862155d68db63be87439e3d68758ad7268e239d39c4fd3b")
     assert P.encode() == exp
 
-testVectorsRistretto(Ed25519Point)
-testVectorsDecaf(Ed448GoldilocksPoint)
-testMapRistretto(Ed25519Point)
-testMapDecaf(Ed448GoldilocksPoint)
+def test():
```
armfazh, line 417: test is not called

armfazh (approved):

poc/ristretto_decaf.sage
chris-wood: Yeah, it was noisy output. We should fix how we run all the tests here, include those for dependencies (like h2c).

PR #229, opened 2021-02-20
Drop UnblindedElement output from test vectors.

chris-wood:
I missed this in #227. Thanks, @bytemare!


chris-wood:
@alxdavids can you please review and merge if this work?

poc/oprf.sage
```diff
@@ -205,6 +206,25 @@ class VerifiableClientContext(ClientContext,Verifiable):
...
         return unblinded_elements
 
+    def finalize_batch(self, xs, blinds, evaluated_elements, blinded_elements, proof):
+        assert(len(blinds) == len(evaluated_elements))
+        assert(len(evaluated_elements) == len(blinded_elements))
+
+        unblinded_elements = self.unblind_batch(blinds, evaluated_elements, blinded_elements, proof)
+
+        outputs = []
+        finalizeDST = _as_bytes("VOPRF06-Finalize-") + self.context_string
```
alxdavids, line 216: Nit: should we use a different DST when batching?

alxdavids (approved):
LGTM, one nit around batching, but I don't have a strong opinion either way.

poc/oprf.sage
chris-wood: Since this is just an optimization, I think the output should be the same.

chris-wood:
I [made these changes in CIRCL](https://github.com/cloudflare/circl/pull/217) and they seem to pass. So I'm merging this as-is.

PR #230, opened 2021-02-20
Specify Generate/DeriveKeyPair.

### LINKED ISSUE #228: Consider adding deterministic KeyGen function
We currently assume the existence of a randomized `KeyGen` function for each of the groups we care about. There are some use cases where one might want to deterministically derive a key (pair) from a seed of appropriate entropy. @alxdavids, what do you think about adding this to the spec?

chris-wood:
Closes #228.


draft-irtf-cfrg-voprf.md
```diff
@@ -309,27 +309,17 @@ The fundamental group operation is addition `+` with identity element
...
+itself `r-1` times, this is denoted as `ScalarBaseMult(r)`. The set of
+scalars corresponds to `GF(p)`.
 
 We now detail a number of member functions that can be invoked on a
-prime-order group.
+prime-order group `GG`.
 
-- Order(): Outputs the order of GG (i.e. `p`).
+- Order(): Outputs the order of `GG` (i.e. `p`).
 - Identity(): Outputs the identity element of the group (i.e. `I`).
```
armfazh, line 322:
```suggestion
- Identity(): Outputs the identity element of the group (i.e. `I`).
- Generator(): Outputs the chosen element of order `p` that 
  generates `GG`.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -342,6 +332,26 @@ prime-order group.
...
+
+Using the API of a prime-order group, we assume the existence of a function
+`GenerateKeyPair()` that generates a random private and public key pair
+(`skS`, `pkS`). One possible implementation might be to compute
+`skS = RandomScalar()` and `pkS = ScalarBaseMult(skS)`. We also assume the
+existence of a `DeriveKeyPair(seed)` function that deterministically generates
+a private and public key pair from input `seed`, where `seed` is a random
+byte string that SHOULD have at least `Ns` bytes of entropy.
+`DeriveKeyPair(seed)` computes `skS = HashToScalar(seed)` and
+`pkS = ScalarBaseMult(skS)`.
```
armfazh, line 354:
```suggestion
Using the API of a prime-order group, we assume the existence
of two functions for generating a private key `skS ` and its
associated public key `pkS`. The function GenerateKeyPair() is 
a randomized algorithm, and `DeriveKeyPair(seed)` uses a seed
of at least `Ns` bytes to deterministically generate a key pair.
One recommended implementation of these functions is given.

def GenerateKeyPair():
  skS = RandomScalar()
  pkS = skS*Generator()
  return skS, pkS

def DeriveKeyPair(seed):
  skS = HashToScalar(seed)
  pkS = skS*Generator()
  return skS, pkS
```

armfazh (approved):
I made a couple of suggestions. Maybe need to emphasize the MUST somewhere.

chris-wood:
@armfazh do we need to expose Generator() on the API if we have ScalarBaseMult?

draft-irtf-cfrg-voprf.md
chris-wood: DeriveKeyPair can’t be just a recommendation, it needs to be a concrete specification, in order to be useful. Two different implementations of this functions must produce the same output. (If they didn’t, then we might as well remove the function entirely, since it adds no value, right?)

armfazh: Yes, please edit accordingly.

chris-wood: I think the current text is sufficient. Is there a reason why it doesn’t work?

armfazh:
> @armfazh do we need to expose Generator() on the API if we have ScalarBaseMult?

That's correct, I haven't seen where ScalarBaseMult was defined. Now, in its definition loosely specifies what the base point is.

draft-irtf-cfrg-voprf.md
armfazh: I am not saying yours is incorrect. My intention is to improve visual readability.

chris-wood: Let's keep it as-is for now. The current version seems to communicate the same thing, but with less words. If we decide later it can be improved, let's fix in a future PR.

PR #233, opened 2021-02-20
Fix ristretto255/decaf448 hash-to-scalar.

### LINKED ISSUE #232: ristretto255 and decaf448 hash-to-scalar issues
Currently, the document defines HashToScalar for these groups like so:

```
Use hash_to_field from {{!I-D.irtf-cfrg-hash-to-curve}} using Order() as the prime modulus, with L = 48, and expand_message_xmd with SHA-512.
```

This aligns with HashToScalar for other groups (based on NIST curves). However, it is somewhat inconsistent with the HashToGroup implementation of ristretto255 and decaf448, which basically rely on HKDF and the one-way maps. I *think* the best way to address this is to provide a one-way map in the ristretto255 document, similar to that which, to my knowledge, [is implemented quite widely](https://github.com/gtank/ristretto255/blob/master/scalar.go#L64-L69). 

Edit: Filippo points out that there's text for this (that I just overlooked):

> Given a uniformly distributed 64-byte string b, implementations can obtain a scalar by interpreting the 64-byte string as a 512-bit integer in little-endian order and reducing the integer modulo l, as in [RFC8032].

So we should just use this. :-)

@gtank, @FiloSottile, @armfazh, @kwantam, @isislovecruft: what do you think?

chris-wood:
Closes #232.

cc @gtank, @FiloSottile, @bytemare


bytemare:
All vectors match 👍  (except Decaf448, which I don't test, yet)

alxdavids (approved):
All looks fine to me 👍

armfazh (approved):
lgtm

FiloSottile:
Not familiar with all the VOPRF draft language, but the interaction with the groups LGTM.

PR #237, opened 2021-02-23
Setting a fixed seed to generate same test vectors.

armfazh:
Fixing a seed of the PRNG.


poc/test_oprf.sage
```diff
@@ -43,7 +44,7 @@ class Protocol(object):
         self.suite = suite
         self.mode = mode
 
-        self.seed = os.urandom(suite.group.scalar_byte_length())
+        self.seed = bytes(ZZ(random.getrandbits(8*suite.group.scalar_byte_length())).digits(2**8))
```
chris-wood, line 47: Why not just fix the seed here to a string of length `Ns`?

chris-wood (changes requested):

poc/test_oprf.sage
armfazh: good trick, it's refactored now

poc/test_oprf.sage
```diff
@@ -43,7 +43,7 @@ class Protocol(object):
         self.suite = suite
         self.mode = mode
 
-        self.seed = os.urandom(suite.group.scalar_byte_length())
+        self.seed = b'\xA3'*suite.group.scalar_byte_length()
```
chris-wood, line 46:
```suggestion
        self.seed = b'\xA3' * suite.group.scalar_byte_length()
```

Issue #234, opened 2021-02-21
Change SHA-512 to SHAKE-256 for Decaf


claucece:
This is basically an extension of #200 

ed448 is implemented with SHAKE-256, as defined and analysed here: https://tools.ietf.org/html/rfc8032#section-8.9

Following the same comment by @jedisct1 , decaf448 is a natural extension to ed448 (even though it is not widely implemented), and it uses shake-256. 

This, of course, mean that the  expand_message function will be different.

I raised this point and #200 when implementing ristretto and decaf in the poc. 

Just wondering the thoughts after #200 went through.

cc./ @jedisct1 @chris-wood

chris-wood:
I don't think we should make this change. SHAKE is not widely supported in cryptographic libraries.

claucece:
The same way that curve448 or decaf is not widely implemented. But when you do implement it for signing (for ed448, which is most of the times what you use the functionality of decaf for), you have to have shake256. So implementing that one sort of implies the other.

chris-wood:
> The same way that curve448 or decaf is not widely implemented. But when you do implement it for signing (for ed448, which is most of the times what you use the functionality of decaf for), you have to have shake256. So implementing that one sort of implies the other.

Yeah, that's true, if you're using ed448 for signing, but it's not clear to me that's a sought-after use case. I'd like to see more widespread support before we force SHAKE onto libraries.

chris-wood:
@gtank, @filosottile: what's your take here?

FiloSottile:
The r255-d448 draft doesn't depend on any hash functions, specifically to leave these decisions to higher-level specs that can aim for consistency. (I find the EdDSA specs introducing a choice of hash without actually having a requirement for a specific one an anti-pattern, for example.) I would recommend picking whatever is consistent within the context of your spec, but I don't have an opinion on what that is.

chris-wood:
> I would recommend picking whatever is consistent within the context of your spec, but I don't have an opinion on what that is.

Yep, and it seems both would be consistent (for some definition of consistency). My take right now is that ease-of-implementation is a primary consideration. If we think most implementations of d448 will come packaged up with ed448 and EdDSA, then perhaps SHAKE is the right answer (since it adds no new dependency). But if we think most implementations of d448 may come with just ed448, and no corresponding EdDSA implementation, then maybe SHA512 is better. I don't know! 🤷

jedisct1:
[...]

PR #238, opened 2021-02-23
Change decaf to use shake #234

claucece:


draft-irtf-cfrg-voprf.md
```diff
@@ -1015,10 +1015,10 @@ and ristretto255. See {{cryptanalysis}} for related discussion.
...
     {{!I-D.irtf-cfrg-hash-to-curve}} with DST =
-    "VOPRF06-HashToGroup-" || contextString, and `expand_message` = `expand_message_xmd`
-    using SHA-512.
-  - HashToScalar(): Compute `uniform_bytes` using `expand_message` = `expand_message_xmd`,
-    DST = "VOPRF06-HashToScalar-" || contextString, and output length 64, interpret
+    "VOPRF06-HashToGroup-" || contextString, and `expand_message` = `expand_message_xof`
+    using SHAKE-256.
+  - HashToScalar(): Compute `uniform_bytes` using `expand_message` = `expand_message_xof`,
+    DST = "VOPRF06-HashToScalar-" || contextString, and output length 112, interpret
     `uniform_bytes` as a 512-bit integer in little-endian order, and reduce the integer
```
armfazh, line 1022:
```suggestion
    `uniform_bytes` as a 896-bit integer in little-endian order, and reduce the integer
```

armfazh (approved):

draft-irtf-cfrg-voprf.md
claucece: ah yes! I was going to check that! Thank you for this! ;)

draft-irtf-cfrg-voprf.md
```diff
@@ -1015,11 +1015,11 @@ and ristretto255. See {{cryptanalysis}} for related discussion.
...
-    "VOPRF06-HashToGroup-" || contextString, and `expand_message` = `expand_message_xmd`
-    using SHA-512.
-  - HashToScalar(): Compute `uniform_bytes` using `expand_message` = `expand_message_xmd`,
-    DST = "VOPRF06-HashToScalar-" || contextString, and output length 64, interpret
-    `uniform_bytes` as a 512-bit integer in little-endian order, and reduce the integer
+    "VOPRF06-HashToGroup-" || contextString, and `expand_message` = `expand_message_xof`
+    using SHAKE-256.
+  - HashToScalar(): Compute `uniform_bytes` using `expand_message` = `expand_message_xof`,
+    DST = "VOPRF06-HashToScalar-" || contextString, and output length 112, interpret
+    `uniform_bytes` as a 896-bit integer in little-endian order, and reduce the integer
```
chris-wood, line 1022: Hmm, why do we need longer output here? (The output lengths previously seemed to align with the conventional way of mapping bytes to scalars in d448 draft.)

poc/ristretto_decaf.sage
```diff
@@ -196,7 +196,7 @@ class DecafPoint(QuotientEdwardsPoint):
     def hash_to_scalar(self, msg, dst=""):
         # this is taking the edwards parameters defined in draft-irtf-cfrg-hash-to-curve
         # hash_to_field(msg, count, dst, modulus, degree (m), blen (l), expand_fn, hash_fn, security_param):
-        return hash_to_field(msg, 1, dst, self.order, 1, 84, expand_message_xmd, hashlib.sha512, 224)[0][0]
+        return hash_to_field(msg, 1, dst, self.order, 1, 84, expand_message_xof, hashlib.shake_256, 224)[0][0]
```
chris-wood, line 199: This doesn't seem right. Shouldn't it be something like this?

```
L = 64 # see comment above!
uniform_bytes = expand_message_xof(msg, dst, L, hashlib.shake_256, _)
return OS2IP_le(uniform_bytes) % self.order
```

chris-wood (changes requested):

draft-irtf-cfrg-voprf.md
claucece: Sure. I can change it back to 64.

chris-wood: Was there a reason you made it 112? (I am genuinely curious!)

poc/ristretto_decaf.sage
claucece: I don't really understand what you are referring to.

L here is

```
L = ceil((ceil(log2(p)) + k) / 8)
```

```
To control bias, hash_to_field instead uses random integers whose length is at least ceil(log2(p)) + k bits, where k is the target security level for the suite in bits.
```

And as defined per the draft, the security level of curve448 is 224. As defined per draft, `L: 84`. 

Is that what this is referring to?

claucece:
```
def hash_to_field(msg, count, dst, modulus, degree, blen, expand_fn, hash_fn, security_param):
    len_in_bytes = count * degree * blen
    uniform_bytes = expand_fn(msg, dst, len_in_bytes, hash_fn, security_param)
    u_vals = [None] * count
    for i in xrange(0, count):
        e_vals = [None] * degree
        for j in xrange(0, degree):
            elm_offset = blen * (j + i * degree)
            tv = uniform_bytes[elm_offset : (elm_offset + blen)]
            e_vals[j] = OS2IP(tv) % modulus
        u_vals[i] = e_vals
    return u_vals
```

which translates to the draft

```
hash_to_field(msg, count)
Parameters:
- DST, a domain separation tag (see discussion above).
- F, a finite field of characteristic p and order q = p^m.
- p, the characteristic of F (see immediately above).
- m, the extension degree of F, m >= 1 (see immediately above).
- L = ceil((ceil(log2(p)) + k) / 8), where k is the security
parameter of the suite (e.g., k = 128).
- expand_message, a function that expands a byte string and
domain separation tag into a uniformly random byte string
(see discussion above).
Inputs:
- msg, a byte string containing the message to hash.
- count, the number of elements of F to output.
Outputs:
- (u_0, ..., u_(count - 1)), a list of field elements.
Steps:
1. len_in_bytes = count * m * L
2. uniform_bytes = expand_message(msg, DST, len_in_bytes)
3. for i in (0, ..., count - 1):
4. for j in (0, ..., m - 1):
5. elm_offset = L * (j + i * m)
6. tv = substr(uniform_bytes, elm_offset, L)
7. e_j = OS2IP(tv) mod p
8. u_i = (e_0, ..., e_(m - 1))
9. return (u_0, ..., u_(count - 1))
```

chris-wood: Sorry, sorry, I just used L for a variable name. It's not meant to refer to *that* L.

chris-wood: (I updated the suggestion to remove L and avoid confusion)

chris-wood:
Taking a step back, how about we specify HashToScalar following the advice in hash-to-curve like so:

*ristretto255:*
```
HashToScalar(): Use hash_to_field from {{!I-D.irtf-cfrg-hash-to-curve}} using Order() as the prime modulus, L = 48, `expand_message_xof` with SHA-512, and DST = "VOPRF06-HashToScalar-" || contextString.
```

*decaf448:* 
```
HashToScalar(): Use hash_to_field from {{!I-D.irtf-cfrg-hash-to-curve}} using Order() as the prime modulus, L = 72, `expand_message_xof` with SHA-512, and DST = "VOPRF06-HashToScalar-" || contextString.
```

The values of L were derived from `L = ceil((ceil(log2(p)) + k) / 8)`, where `p = Order()` and `k = 128` (for a 128-bit security level). That would make each of these consistent with the other HashToScalar invocations. 

@armfazh, thoughts?

bytemare:
If SHA-512, it would be `expand_message_xmd` (not xof)

bytemare:
Yes, and I also meant for Ristretto255 :)

chris-wood:
> Yes, and I also meant for Ristretto255 ;)

We discussed this in #200. Basically, it makes most sense to use SHA-512 and `_xmd` for r225.

bytemare:
I know, but you wrote `_xof` on the line for SHA-512/r255 :)

armfazh:
> Taking a step back, how about we specify HashToScalar following the advice in hash-to-curve like so:
> That would make each of these consistent with the other HashToScalar invocations.

this seems reasonable to me, one single hashToField implementation can serve to all the instantiations, if code reuse is a goal.

claucece:
@bytemare noted ;) I think we all agree on that. On the pr, it will be expand_md with sha512 for r255; and expand_xof with shake256 for d448.

claucece:
I like the last proposal @chris-wood . I'll change the need text ;) Thanks!

claucece:
@chris-wood @armfazh let me know what you think now ;)

chris-wood (approved):
Looks good!

armfazh:
> let me know what you think now ;)

good,

chris-wood:
@alxdavids are you OK with this? If so, let's merge it!

tmthrgd:
As someone coming across and implementing this spec, I'm not a fan of this change in general. (I'll try and elaborate elsewhere). But specifically as it stands this change seems really odd to me. It replaces `expand_message_xmd` with SHA-512  with `expand_message_xof` with SHAKE-256, but it doesn't change Hash (required for finalize, etc.) from SHA-512.

Far from removing a potential dependency this just adds another. OPRF(decaf448, SHA-512) would require *both* SHA-512 *and* SHAKE-256 to implement. That's the worst of both worlds. If there is a desire to do this, surely Hash should also be swapped out for SHA3/SHAKE?

chris-wood:
> Far from removing a potential dependency this just adds another. OPRF(decaf448, SHA-512) would require both SHA-512 and SHAKE-256 to implement. That's the worst of both worlds. If there is a desire to do this, surely Hash should also be swapped out for SHA3/SHAKE?

Hah, good catch :-) I believe the *intent* was to replace all instances of SHA-2 with SHAKE for decaf. I think we all just overlooked the corresponding hash.

claucece:
@tmthrgd indeed, it will be changed as well. 

> would require both SHA-512 and SHAKE-256 to implement

Generally, SHA-512 is widely implemented and in most languages that have the SHA-3 family, SHA-2 is also available. 

Is there a language that has SHA-3 family but not SHA-2?

I'll change the pr to have SHAKE eveywhere.

claucece:
@tmthrgd should be changed now ;)

chris-wood:
@claucece any updates on this?

claucece:
@chris-wood email was sent asking for the clarifications that you needed. Last time I checked, this was not a big priority prior to IETF111.. is there a reason to make this a bigger priority now? Thanks!

chris-wood:
> is there a reason to make this a bigger priority now? Thanks!

Nope! I'm just curious to know what the status is for this change.

chris-wood:
Closing in favor of #249.

PR #239, opened 2021-03-01
Generalize GenerateProof/VerifyProof

chris-wood:
This will let folks re-use these functions more generically if they so choose.

cc @siyengar, @ananthr

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -227,7 +227,7 @@ security properties.
 - Fold Unblind function into Finalize.
 - Optimize ComputeComposites for servers (using knowledge of the private key).
 - Specify deterministic key generation method.
-- Update test vectors. 
+- Update test vectors.
 - Apply various editorial changes.
 
 [draft-05](https://tools.ietf.org/html/draft-irtf-cfrg-voprf-05):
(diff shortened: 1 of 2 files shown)


alxdavids:
It looks like this is re-introducing group elements to the public API, is this a good idea? It doesn't seem to match the rest of the VOPRF API (non-proof-related functions) which operates over bytes.

chris-wood:
> It looks like this is re-introducing group elements to the public API, is this a good idea? It doesn't seem to match the rest of the VOPRF API (non-proof-related functions) which operates over bytes.

Hmm, these are all internal functions, right? Since they're internal, making them operate on elements rather than byte strings seems perfectly fine. (An implementation ought to know how to represent these types.) 

As I see it, the public APIs only include Blind, Evaluate, and Finalize, and those all operate on "public" types like SerializedElements.

alxdavids:
> > It looks like this is re-introducing group elements to the public API, is this a good idea? It doesn't seem to match the rest of the VOPRF API (non-proof-related functions) which operates over bytes.
> 
> Hmm, these are all internal functions, right? Since they're internal, making them operate on elements rather than byte strings seems fine. (An implementation ought to know how to represent and these types locally.)
> 
> As I see it, the public APIs only include Blind, Evaluate, and Finalize, and those all operate on "public" types like SerializedElements that may be received over the wire.

Okay, yeah that makes sense, I was confused. In that case, I think the changes are probably fine. Out of interest, what was the main sticking point with the previous version of the API?

alxdavids (approved):

chris-wood:
> Okay, yeah that makes sense, I was confused. In that case, I think the changes are probably fine. Out of interest, what was the main sticking point with the previous version of the API?

It wasn't usable by Subodh's draft, which led them to define (yet another) DLEQ proof generation and verification routine. I threw up a [PR based on the changes here](https://github.com/siyengar/voprf-with-metadata/pull/1) that makes these two work really well together.

siyengar:
This looks great. I think the only difference between this and the DLEQ proofs in the other draft at present is that this one uses subtraction for the proofs and the other one uses addition

chris-wood:
> This looks great. I think the only difference between this and the DLEQ proofs in the other draft at present is that this one uses subtraction for the proofs and the other one uses addition

Would you prefer one versus the other? :-)

siyengar (approved):

siyengar:
> > This looks great. I think the only difference between this and the DLEQ proofs in the other draft at present is that this one uses subtraction for the proofs and the other one uses addition
> 
> Would you prefer one versus the other? :-)

na, don't care

armfazh (approved):

Issue #225, opened 2021-02-05
Consider adding no-DLEQ ciphersuite for VOPRF


alxdavids:
The PMB Tokens paper constructs a weaker version of a VOPRF without DLEQ proofs (see Section 7 of https://eprint.iacr.org/2020/072). The weakness of the construction allows the server to send back random garbage and the client cannot distinguish.

This effectively provides the server with an extra bit of privacy that it can remove. However, removing the DLEQ proofs also comes with the benefit of removing some of the more expensive computation from the server, so I think it should be considered as a separate ciphersuite. Note, this is only valid for the VOPRF in the additive blinding case.

First raised by @siyengar. cc @chris-wood, @armfazh, @claucece for further discussion.

bytemare:
I may be misunderstanding, but OPAQUE is defined to use the non-verifiable OPRF, so already without any specific overhead. What would be the advantage ?

chris-wood:
Given that this possibly worsens the privacy posture of the protocol and is limited to a particular implementation of the blinding step, I'm not inclined to make this change. @siyengar, can you elaborate on the benefit here? (I probably just didn't understand the motivation.)

siyengar:
Ya it likely does worsen the privacy posture, although this is probably a bit more complicated as there are some issues with the current scheme as well if the client fails to perform proper validation. My thought on the benefits of this type of scheme were 3 fold:

1. Gets rid of the client needing to perform a validation step. As we know from TLS and other protocols, adding an `if` condition to the protocol allows people not to execute that `if` condition. And if people don't execute the proof then that also reduces the privacy of the protocol.
2. Reduces number of bytes needed to be downloaded from the server. You don't need to download an extra dleq proof step with this. 
3. A More intuitive protocol. You get a token and then replay the token, so it becomes an intuitively simpler protocol.

chris-wood:
I don't disagree with any of the points, but I really don't understand the proposal. Cold you or @alxdavids whip up a draft PR with something more concrete?

chris-wood:
Okay, I RTFM :-) this seems less like a new ciphersuite and more like a new mode because (a) the group doesn't matter, (b) implementation details differ compared to the other two modes, and (c) the security properties are different. Perhaps we could call this "WeakVerifiable" mode, or something? It might look like so:

1. Blind is as specified in Section 7 of [the PMB paper](https://eprint.iacr.org/2020/072.pdf).
[...]

PR #240, opened 2021-03-02
Add draft of WeakVerifiable mode.

chris-wood:
First attempt at #225. Opening as a draft while we work through some key questions:

1. The protocol sketch in the PMBTokens paper suggests that the server produces a DLEQ proof of its public key, and that the client verifies this before running the protocol. However, it's not clear to me what value this check adds. There seem to be two cases: (a) the key is invalid, in which case the client doesn't use it, or (b) the key is valid, in which case the client uses it (but is possible put into two buckets based on the server response). Is it necessary that the server produce a proof *before* the online protocol is run? @dvorak42, do you know? Could you please check with the authors? :-) 
2. Given (1), the proposed change right now re-uses the server's base context in the weak verifiable mode. This is because the server produces no DLEQ proof online, for each evaluation. 

cc @siyengar, @ananthr


dvorak42:
I'll double-check with the authors tomorrow, but I believe the proof is proving that the server is capable of signing the token correctly. I don't think this is particularly necessary ahead of the protocol for this use case since for a server that can't sign the token correctly, what ends up out of the system is random garbage, similar to the case where the server multiples randomly.

chris-wood:
Yeah, that matches my understanding. Thanks @dvorak42 :-)

PR #243, opened 2021-03-31
Refactor domain separation strings to allow for general use.

chris-wood:
The current domain separation strategy prepended a variant and version string to each internal DST. If the proof generation and verification routines were reused elsewhere, e.g., in other standards, that would mean the DST would include "VOPRF," which is not ideal. (Certainly this proof logic is not coupled to the rest of the VOPRF protocol!) This change moves the variant and version string around to the context string, and makes that a parameter for each function. If another standard wants to change the string, they can do so by changing the context string parameter to whatever suits their needs. 

cc @siyengar


draft-irtf-cfrg-voprf.md
```diff
@@ -505,7 +507,11 @@ SerializedScalar Proof[2];
 ## Context APIs {#api}
 
 In this section, we detail the APIs available on the client and server
-(V)OPRF contexts.
+(V)OPRF contexts. Each API has the following implicit parameters:
+
+- GG, a prime-order group implementing the API described below in {{pog}}.
```
chris-wood, line 512:
```suggestion
- GG, a prime-order group implementing the API described in {{pog}}.
```

armfazh (approved):

PR #246, opened 2021-04-30
Highlighting key generation functions.

armfazh:
This change unhides the key related operations based on primer order groups.


draft-irtf-cfrg-voprf.md
```diff
@@ -354,15 +345,30 @@ prime-order group `GG`.
...
+  defined in {{!I-D.irtf-cfrg-hash-to-curve}} setting the target field to GF(p).
+- GenerateKeyPair(): Samples at random a private key `skS`, generates
+  the corresponding public key `pkS`, and returns this key pair. One
+  possible implementation might be to compute `skS = RandomScalar()`
+  and `pkS = ScalarBaseMult(skS)`.
+- DeriveKeyPair(seed): Deterministically generates a private key from a
+  `seed`, which is a random byte string that SHOULD have at least `Ns`
+  bytes of length. Then, it generates the corresponding public key `pkS`,
+  and returns this key pair. One possible implementation might be to
+  compute `skS = HashToScalar(seed)` and `pkS = ScalarBaseMult(skS)`.
```
chris-wood, line 371: Why not make this the required implementation? (OPAQUE, for example, depends on the function being implemented this way.)

draft-irtf-cfrg-voprf.md
```diff
@@ -326,21 +326,12 @@ application of the group operation `G` with itself `r-1` times, this
 is denoted as `ScalarBaseMult(r)`. The set of scalars corresponds to
 `GF(p)`.
 
+### Mathematical Operations
```
chris-wood, line 329:
```suggestion
```

draft-irtf-cfrg-voprf.md
```diff
@@ -354,15 +345,30 @@ prime-order group `GG`.
...
-Using the API of a prime-order group, we assume the existence of a function
-`GenerateKeyPair()` that generates a random private and public key pair
-(`skS`, `pkS`). One possible implementation might be to compute
-`skS = RandomScalar()` and `pkS = ScalarBaseMult(skS)`. We also assume the
-existence of a `DeriveKeyPair(seed)` function that deterministically generates
-a private and public key pair from input `seed`, where `seed` is a random
-byte string that SHOULD have at least `Ns` bytes of entropy.
-`DeriveKeyPair(seed)` computes `skS = HashToScalar(seed)` and
-`pkS = ScalarBaseMult(skS)`.
+### Cryptographic Operations
```
chris-wood, line 348:
```suggestion
```

draft-irtf-cfrg-voprf.md
```diff
@@ -354,15 +345,30 @@ prime-order group `GG`.
...
+
+Using the API of a prime-order group, we describe the some functions used for
+cryptographic purposes.
+
+- HashToGroup(x): A member function of `GG` that deterministically maps
```
chris-wood, line 353: Hmm, why isn't this an API on the group? (The description specifically says this is a member function of `GG`, for example.)

draft-irtf-cfrg-voprf.md
```diff
@@ -354,15 +345,30 @@ prime-order group `GG`.
...
+  an array of bytes `x` to an element of `GG`. The map must ensure that,
+  for any adversary receiving `R = HashToGroup(x)`, it is
+  computationally difficult to reverse the mapping. Examples of hash to
+  group functions satisfying this property are described for prime-order
+  (sub)groups of elliptic curves, see {{!I-D.irtf-cfrg-hash-to-curve}}.
+- HashToScalar(x): A member function of `GG` that deterministically maps
+  an array of bytes `x` to an element in GF(p). A recommended method
+  for its implementation is instantiating the hash to field function,
+  defined in {{!I-D.irtf-cfrg-hash-to-curve}} setting the target field to GF(p).
+- GenerateKeyPair(): Samples at random a private key `skS`, generates
```
chris-wood, line 363: GenerateKeyPair and DeriveKeyPair seem to be the two functions here that are probably best pulled out of the POG API. What do you think?

draft-irtf-cfrg-voprf.md
```diff
@@ -354,15 +345,30 @@ prime-order group `GG`.
...
-`skS = RandomScalar()` and `pkS = ScalarBaseMult(skS)`. We also assume the
-existence of a `DeriveKeyPair(seed)` function that deterministically generates
-a private and public key pair from input `seed`, where `seed` is a random
-byte string that SHOULD have at least `Ns` bytes of entropy.
-`DeriveKeyPair(seed)` computes `skS = HashToScalar(seed)` and
-`pkS = ScalarBaseMult(skS)`.
+### Cryptographic Operations
+
+Using the API of a prime-order group, we describe the some functions used for
+cryptographic purposes.
```
chris-wood, line 351:
```suggestion
We also assume the two following functions for generating OPRF key pairs.
```

draft-irtf-cfrg-voprf.md
armfazh: I agree, so we can relocate them to a different section. (together with the hash* see other comment)

draft-irtf-cfrg-voprf.md
armfazh: it's true, it doesn't need to be part of "the crypto API"

armfazh:
ok, I'm going to refactor based on these comments.

armfazh:
I've updated the PR according to the review.

draft-irtf-cfrg-voprf.md
```diff
@@ -446,13 +436,37 @@ blinded token. In `Finalize` the client unblinds the server response,
 verifies the server's proof if verifiability is required, and produces
 a byte array corresponding to the output of the OPRF protocol.
 
+## Key Generation
```
chris-wood, line 439: Can we move this section up to the where the old text was? It seems more natural to follow the description of the POG.

draft-irtf-cfrg-voprf.md
```diff
@@ -446,13 +436,37 @@ blinded token. In `Finalize` the client unblinds the server response,
...
+
+Two functions are defined for generating a (V)OPRF key pair (`skS`, `pkS`).
+The `GenerateKeyPair` function samples at random a private key `skS`,
+whereas the `DeriveKeyPair` function deterministically generates a
+private key from a random byte string `seed` that SHOULD have at least
+`Ns` bytes of length. Once the private key was obtained, both functions
+generate the public key `pkS` in the same way.
+
+~~~
+def GenerateKeyPair(suite):
```
chris-wood, line 449: Hmm, I think we should only write a description for DeriveKeyPair, and then just list requirements for GenerateKeyPair. Some folks may want to implement GenerateKeyPair a different way, e.g., due to FIPS reasons or something, and we don't want to prohibit that.

chris-wood (changes requested):

draft-irtf-cfrg-voprf.md
armfazh: take a look on latest changes.

draft-irtf-cfrg-voprf.md
```diff
@@ -354,15 +354,15 @@ prime-order group `GG`.
...
-`pkS = ScalarBaseMult(skS)`.
+Two functions can be used for generating a (V)OPRF key pair (`skS`, `pkS`).
+The `DeriveKeyPair` function deterministically generates the private key
+`skS` given a random byte string `seed` that SHOULD have at least `Ns`
+bytes of length. To do that, compute `skS = GG.HashToScalar(seed)`.
+Then, the public key is generated as `pkS = ScalarBaseMult(skS)`.
+The `GenerateKeyPair` function is a randomized algorithm that outputs a
+key pair (`skS`, `pkS`). Both functions must guarantee that the private
+key `skS` is a non-zero integer less than `p`, and the public key is
+generated as `pkS = ScalarBaseMult(skS)`.
```
chris-wood, line 365: Why does `pkS` need to be _generated_ with this function? Maybe we can just say that `pkS = ScalarBaseMult(skS)`? (The difference being that the latter specifies the relation between `skS` and `pkS`, rather than exactly how one computes `pkS` given `skS`.)

armfazh, line 365: Better now?  (feel free to edit)

```suggestion
Two functions can be used for generating a (V)OPRF key pair (`skS`, `pkS`).
The `DeriveKeyPair` function deterministically generates the private key
`skS` given a random byte string `seed` that SHOULD have at least `Ns`
bytes of length; to do that, compute `skS = GG.HashToScalar(seed)`.
The `GenerateKeyPair` function is a randomized algorithm that outputs a
key pair (`skS`, `pkS`). Both functions must guarantee that the private
key `skS` is a non-zero integer less than `p`, and the public key is
`pkS = ScalarBaseMult(skS)`.
```

chris-wood, line 365: How about this slightly different variant?

```suggestion
Two functions can be used for generating a (V)OPRF key pair (`skS`, `pkS`)
with the relation `pkS = ScalarBaseMult(skS)`: `GenerateKeyPair` and 
`DeriveKeyPair`. `GenerateKeyPair` is a randomized function that outputs
a fresh key pair (`skS`, `pkS`) upon ever invocation. `DeriveKeyPair` is a 
deterministic  function that generates private key `skS` from a random byte 
string `seed` that  SHOULD have at least `Ns` bytes of entropy, and then 
computes `pkS = ScalarBaseMult(skS)`. Both functions must guarantee that 
the private key `skS` is a non-zero integer less than `p`.
```

armfazh: so much better,

chris-wood (approved):

Issue #236, opened 2021-02-23
Deriving keys deterministically requires domain separation?


armfazh:
In this call to HashToScalar, no DST is specified, which is not conformant with Ciphersuite section in the draft.

https://github.com/cfrg/draft-irtf-cfrg-voprf/blob/866a54cc1021390359e67d8a1b773e2d70f19067/poc/oprf.sage#L281-L284

OTOH, setting `DST = "VOPRF-HashToScalar"|| context_string` is a dst specific to the (v)oprf protocol. It seems to me that the derivation of keys must be agnostic of the caller protocol.

tmthrgd:
For reference: changing `HashToScalar` to incorporate DST was part of #233.

chris-wood:
@armfazh is there a reason why we'd want the key derivation procedure to be decoupled from the protocol in this specification? To me, coupling them seems attractive, as I can imagine applications being tempted to re-use seeds for multiple protocols. This coupling would do its best to ensure that the key output from `DeriveKeyPair` is different from other keys derived from that seed. (Of course, applications could just reuse the derived keys in multiple protocols, but there's not much we can do about that from a specification level.)

@alxdavids, @claucece, what do you think?

armfazh:
My main concern is that DeriveKeyPair, as is currently specified in the draft, uses `HashToScalar(seed, dst)`, with `dst="VOPRF..."`.  But, the reference implementation sets `dst` to the empty string. So this needs to be fixed.

Also, I was slighly on the side of independent key generation, but I don't really have a strong preference about whether deriving keys must be coupled or not to the protocol.

chris-wood:
> My main concern is that DeriveKeyPair, as is currently specified in the draft, uses HashToScalar(seed, dst), with dst="VOPRF...". But, the reference implementation sets dst to the empty string. So this needs to be fixed.

Yeah, we can fix the reference implementation if this is the case.

armfazh:
>> armfazh: My main concern is that DeriveKeyPair, as is currently specified in the draft, uses HashToScalar(seed, dst), with dst="VOPRF...". But, the reference implementation sets dst to the empty string. So this needs to be fixed.

> chris-wood: Yeah, we can fix the reference implementation if this is the case.

While trying to patch deriveKeyPair function as follows

```patch 
def DeriveKeyPair(suite, seed):
[...]

PR #247, opened 2021-05-01
Use a per-mode-ciphersuite DST for deterministic key derivation

armfazh:
no updates in the spec, this updates the code.

Adds the version as prefix in several parts.


chris-wood:
#243 changed the domain separation strings such that they're of the following general form:

```
LABEL-STRING || VERSION || MODE || SUITE-ID
```

So, for example, the finalize DST is:

```
"Finalize-" || "VOPRF06-" || MODE || SUITE-ID
```

This change seems to make the code do something like:

```
"VOPRF06-" || "Finalize-" || MODE || SUITE-ID
```

Which is a spec change. Was that intended? (For context, the change in #243 was to allow better re-use of internal functions for the DLEQ proofs _outside_ of this spec.)

armfazh:
> chris-wood:
> #243 changed the domain separation strings such that they're of the following general form:
> Which is a spec change. Was that intended? 

No, I did changes in a old version. I fix that now.
The only "breaking change" is on `DeriveKeyPair`, now it uses the DST dervied per-mode and per-ciphersuite. (This was commented in #236 )

poc/oprf.sage
```diff
@@ -289,31 +293,29 @@ class VerifiableServerContext(ServerContext,Verifiable):
 
 mode_base = 0x00
```
chris-wood, line 294: Unrelated, but should we also make these constant names all-caps?

chris-wood (approved):

chris-wood:
@armfazh can we file an issue to track updating this in CIRCL?

poc/oprf.sage
armfazh: done

Issue #242, opened 2021-03-17
NIST curve scalar serialization


chris-wood:
We currently say that implementations should reduce mod `p`, but `p` is elsewhere the order of the group (and not the underlying field). We should probably fix the language here.

chris-wood:
Fixed in #249.

PR #249, opened 2021-06-08
Adopt SHAKE-256 for decaf448

### LINKED ISSUE #234: Change SHA-512 to SHAKE-256 for Decaf

chris-wood:
This also fixes some serialization ambiguities.

Closes #234, #242.

cc @claucece


armfazh (approved):

draft-irtf-cfrg-voprf.md
```diff
@@ -1020,27 +1022,27 @@ and ristretto255. See {{cryptanalysis}} for related discussion.
...
     bytes. For group elements, use the 'Encode' and 'Decode' functions from
-    {{!RISTRETTO}}. For scalars, ensure they are fully reduced modulo p and
-    in little-endian order.
+    {{!RISTRETTO}}. For scalars, ensure they are fully reduced modulo `Order()`
+    and in little-endian order.
 - Hash: SHA-512, and Nh = 64.
 - ID: 0x0001
 
-## OPRF(decaf448, SHA-512)
+## OPRF(decaf448, SHAKE-256)
```
alxdavids, line 1030: I haven't really been involved in the SHAKE discussions, but is there a reason why SHAKE is suitable for decaf, but not for ristretto?

alxdavids (approved):
Looks fine to me!

draft-irtf-cfrg-voprf.md
bytemare: It's about the consistency with regard to their original definitions, cf #234

PR #250, opened 2021-06-08
Pin blinding mechanism to mode

### LINKED ISSUE #241: Blinding blues
[Jarecki et al.](https://eprint.iacr.org/2021/273) have an upcoming paper at [PKCS2021](https://pkc.iacr.org/2021) that analyzes security properties of blinding mechanisms used in the 2HashDH construction. Recall, for the purposes of this issue, there are two types of blinding (using group notation to align with this specification, whereas their paper uses exponential notaton):

- Multiplicative: 
   - C -> S: a = rH’(x), for random scalar r
   - S -> C: b = ak, for server private key k
   - C: output H(x, br^(-1)) = H(x, kH’(x))
- Additive:
   - C -> S: a = H’(x) + rG, for random scalar r and fixed generator G
   - S -> C: (b = ak, z = kG), for server private key k
   - C: output H(x, b - zr) = H(x, H’(x))

A generalized summary of their results, without accounting for particular application properties, is as follows.

- Multiplicative blinding is safe. 
- Additive blinding is possibly unsafe, unless:
   - The client has a certified copy of the server public key. This always applies in the verifiable mode by necessity. There may be cases where the client has a certified copy of the key in the base mode.
   - The client input is high entropy, e.g., in the case of Privacy Pass, 32 random bytes of data.
   - The client mixes the public key z = kG into the OPRF evaluation, e.g., by computing H(x, z, kH’(x)) instead of H(x, kH’(x)).

The fundamental problem is as follows: additive blinding with a maliciously created z, i.e., one different than kG, gives the attacker a way of testing one input per OPRF interaction. Applications wherein this is possible regardless of the blinding mechanism, e.g., OPAQUE, are not affected (additive blinding is OK). However, for applications where this attacker capability  does not otherwise exist, additive introduces a real weakness.

All in all, this means the choice of blinding mechanism has security implications. The draft should ideally offer a sane default for applications that know what they’re doing (see similar text in [hash-to-curve](https://cfrg.github.io/draft-irtf-cfrg-hash-to-curve/draft-irtf-cfrg-hash-to-curve.html#section-10)), with options and guidance for applications in doing something different if their circumstances warrant it. In considering these defaults, there are a number of options on the table, accounting for code reuse, performance, and bandwidth. 

Here’s what I propose we do to address this issue.

First, refactor the document slightly to permit different types of Blind and Unblind implementations. For example, Blind might be implemented using multiplicative, additive, or both types of blinding. While doing this, require that additive blinding always include the server public key in the Unblind output so that it’s folded into the Finalize computation. This would promote both types of blindings to the main part of the document, whereas additive blinding is currently specified in an appendix.

Second, map each mode to a particular implementation in the following way:

- Verified mode uses additive blinding, by default, since it is strictly a performance improvement and the client is expected to have a certified copy of the public key.
- Weak verifiable mode (see [issue #225](https://github.com/cfrg/draft-irtf-cfrg-voprf/issues/225) uses both additive and multiplicative.
- Base mode uses multiplicative blinding, by default, as it does not require the server public key. Elsewhere, perhaps in an appendix, we can clarify that applications can use additive blinding only if the server public key is available or if the client input is high entropy.

chris-wood:
This change does three things:

1. Mandates additive blinding for verifiable mode.
2. RECOMMENDS multiplicative blinding for base mode.
3. Adds security considerations for applications discussing when it's appropriate to use additive blinding in base mode.

Closes #241.

cc @hugokraw for a look at the blind considerations text :-)


draft-irtf-cfrg-voprf.md
```diff
@@ -936,20 +982,39 @@ Input:
...
+  N := Z - blindedPublicKey
   unblindedElement = GG.SerializeElement(N)
 
   return unblindedElement
 ~~~
 
+The internal `VerifiablePreprocess` function computes a blind and the corresponding
+blinded version of the group generator. By default, it takes no inputs. If the server
+public key is available when invoked, it can be supplied here to produce a blinded
+representation of the public key. This avoids computing it in `VerifiableUnblind`.
```
armfazh, line 1003:
```suggestion
representation of the public key. This avoids computing `blindedPublicKey` in `VerifiableUnblind`.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1278,6 +1343,34 @@ instantiations of this functionality based on the functions described in
...
 in {{!I-D.irtf-cfrg-hash-to-curve}} when instantiating the function.
 
+## Blinding Considerations {#blind-considerations}
+
+This document makes use of two types of blinding variants: multiplicative and
+additive. The advantage of additive blinding is that it allows the client to
+pre-process tables of blinded scalar multiplications for the group generator.
+This may give it a computational efficiency advantage (due to the fact that a
+fixed-base multiplication can be calculated faster than a variable-base
+multiplication). Pre-processing also reduces the amount of computation that
```
armfazh, line 1353:
```suggestion
pre-process tables of multiples of the group generator.
This gives computational efficiency advantage, due to the fact that, by
using precomputation, fixed-base multiplications can be calculated faster
than variable-base multiplications. Pre-processing also reduces the
amount of computation that
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1278,6 +1343,34 @@ instantiations of this functionality based on the functions described in
...
+needs to be done in the online exchange.
+
+However, the choice of blinding mechanism has security implications. {{JKX21}}
+analyze the security properties of both blinding mechanisms used in this
+document. The results can be summarized as follows:
+
+- Multiplicative is safe for all applications.
```
armfazh, line 1360:
```suggestion
- Multiplicative blinding is safe for all applications.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1278,6 +1343,34 @@ instantiations of this functionality based on the functions described in
...
+- Additive blinding is possibly unsafe, unless one of the following conditions
+  is met: The client has a certified copy of the server public key, as is the
+  case in verifiable mode, the client input has high entropy, or the client mixes
+  the public key into the OPRF evaluation.
```
armfazh, line 1364:
```suggestion
  is met: The client has a certified copy of the server public key (as is the
  case in verifiable mode), the client input has high entropy, or the client mixes
  the public key into the OPRF evaluation.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1278,6 +1343,34 @@ instantiations of this functionality based on the functions described in
...
+
+To avoid security issues with the base mode, this specification RECOMMENDS use of
+multiplicative blinding. This is because it is not known if the server public key
+is available or if the client input has high entropy. Applications wherein either
+of these conditions are true MAY use additive blinding.
+
+The verifiable mode always makes use of the more optimal additive blinding variant,
```
armfazh, line 1371:
```suggestion
The verifiable mode always makes use of the more efficient additive blinding variant,
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1278,6 +1343,34 @@ instantiations of this functionality based on the functions described in
...
+as the public key is available by necessity for verifying the proof.
```
armfazh, line 1372:
```suggestion
as the public key is always available for verifying the proof.
```

armfazh (approved):
in general is good, just left some ediorial suggestions

draft-irtf-cfrg-voprf.md
```diff
@@ -936,20 +982,39 @@ Input:
...
+  blindedPublicKey = blind * pkS
+  N := Z - blindedPublicKey
   unblindedElement = GG.SerializeElement(N)
 
   return unblindedElement
 ~~~
 
+The internal `VerifiablePreprocess` function computes a blind and the corresponding
+blinded version of the group generator. By default, it takes no inputs. If the server
+public key is available when invoked, it can be supplied here to produce a blinded
```
alxdavids, line 1002: Should we be more specific about how the server key is blinded if it is provided here? Perhaps offering it as an optional input below and then returning either nil (if it wasn't supplied) or the blinded version. I would just be worried that this text leaves it up to interpretation currently.

alxdavids: Perhaps something like this:
```
~~~
Input:
  PublicKey? pkS

Output:
  Element blindedGenerator
  Element blindedPublicKey
  Scalar blind

def Preprocess(pkS):
  blind = GG.RandomScalar()
  blindedGenerator = ScalarBaseMult(blind)
  if (!pkS):
    return blindedGenerator, null, blind
  blindedPublicKey = ScalarMult(pkS, blind)
  return blindedGenerator, blindedPublicKey, blind
~~~
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1278,6 +1343,35 @@ instantiations of this functionality based on the functions described in
 must adhere to the implementation and security considerations discussed
 in {{!I-D.irtf-cfrg-hash-to-curve}} when instantiating the function.
 
+## Blinding Considerations {#blind-considerations}
+
+This document makes use of two types of blinding variants: multiplicative and
+additive. The advantage of additive blinding is that it allows the client to
+pre-process tables of multiples of the group generator.
+This gives computational efficiency advantage, due to the fact that, by
```
alxdavids, line 1351: Maybe we should be less certain here, because the advantage is only provided that the pre-computation is done.

```suggestion
This can provide a computational efficiency advantage due to the fact that, by
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1278,6 +1343,35 @@ instantiations of this functionality based on the functions described in
...
+
+However, the choice of blinding mechanism has security implications. {{JKX21}}
+analyze the security properties of both blinding mechanisms used in this
+document. The results can be summarized as follows:
+
+- Multiplicative blinding is safe for all applications.
+- Additive blinding is possibly unsafe, unless one of the following conditions
+  is met: The client has a certified copy of the server public key (as is the
+  case in verifiable mode), the client input has high entropy, or the client mixes
+  the public key into the OPRF evaluation.
```
alxdavids, line 1365:
```suggestion
  is met: 
    - The client has a certified copy of the server public key (as is the case in the verifiable mode)
    - The client input has high entropy
    - The client mixes the public key into the OPRF evaluation.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1278,6 +1343,35 @@ instantiations of this functionality based on the functions described in
...
+
+To avoid security issues with the base mode, this specification RECOMMENDS use of
```
alxdavids, line 1367:
```suggestion
To avoid security issues with the base mode, where some of the above conditions may not be met, this specification RECOMMENDS use of
```

alxdavids (approved):
I think it looks good, I just have a few nits.

draft-irtf-cfrg-voprf.md
chris-wood: I resolved this slightly differently by using the public key that's available upon context creation.

PR #252, opened 2021-06-16
Add explicit deserialization errors, and describe validation logic.

### LINKED ISSUE #251: Should deserializing to the identity element be considered invalid?
The specification of `RandomScalar` says:

> RandomScalar(): A member function of GG that chooses at random a non-zero element in GF(p).

Which means that `blindedElement` should never be the identity element.

At the moment, the server runs `DeserializeElement(BlindedElement)` as the first step of Evaluate:

> DeserializeElement(buf): A member function of GG that maps a byte array buf to a group element A, or fails if the input is not a valid byte representation of an element.

What if `BlindedElement` is indeed the identity element -- should this be considered invalid and rejected immediately by the server? Or should the server continue processing the request even though the client could not have generated this by following the protocol?

chris-wood:
Closes #251.

cc @kevinlewi, @bytemare

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -1183,7 +1183,7 @@ and ristretto255. See {{cryptanalysis}} for related discussion.
 
 # API Considerations {#apis}
 
-Some VOPRF APIs specified in this document are fallible. For example, `Finalize()`
+Some VOPRF APIs specified in this document are fallible. For example, `Finalize`
 and `Evaluate` can fail if any element received from the peer fails deserialization.
 The explicit errors generated throughout this specification, along with the
 conditions that lead to each error, are as follows:
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -185,6 +185,7 @@ informative:
       -
         org: J. Xu
         name: Jiayu Xu
+  keyagreement: DOI.10.6028/NIST.SP.800-56Ar3
 
 --- abstract
 
@@ -364,12 +365,15 @@ prime-order group `GG`.
   to a unique byte array `buf` of fixed length `Ne`.
 - DeserializeElement(buf): A member function of `GG` that maps a byte array
   `buf` to a group element `A`, or fails if the input is not a valid
-  byte representation of an element.
+  byte representation of an element. This function can raise a
+  DeserializeError if deserialization fails or `A` is the identity element
+  of the group; see {{input-validation}}.
 - SerializeScalar(s): A member function of `GG` that maps a scalar element `s`
   to a unique byte array `buf` of fixed length `Ns`.
 - DeserializeScalar(buf): A member function of `GG` that maps a byte array
   `buf` to a scalar `s`, or fails if the input is not a valid byte
-  representation of a scalar.
+  representation of a scalar. This function can raise a
+  DeserializeError if deserialization fails; see {{input-validation}}.
 
 Two functions can be used for generating a (V)OPRF key pair (`skS`, `pkS`)
 where `skS` is a non-zero integer less than `p` and `pkS = ScalarBaseMult(skS)`:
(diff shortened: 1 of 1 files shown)


bytemare:
Looks good !

draft-irtf-cfrg-voprf.md
```diff
@@ -971,7 +979,9 @@ def VerifiableBlind(input, blindedGenerator):
   return blindedElement
 ~~~
 
-The inverse `VerifiableUnblind` is implemented as follows.
+The inverse `VerifiableUnblind` is implemented as follows. This function
+can raise an exception if element deserialization or proof verification
+fals.
```
armfazh, line 984:
```suggestion
fails.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1169,6 +1181,20 @@ and ristretto255. See {{cryptanalysis}} for related discussion.
 - Hash: SHA-512, and Nh = 64.
 - ID: 0x0005
 
+# API Considerations {#apis}
+
+Some VOPRF APIs specified in this document are fallible. For example, `Finalize()`
+and `Evaluate` can fail any element received from the peer fails deserialization.
```
armfazh, line 1187:
```suggestion
and `Evaluate` can fail if any element received from the peer fails deserialization.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1169,6 +1181,20 @@ and ristretto255. See {{cryptanalysis}} for related discussion.
 - Hash: SHA-512, and Nh = 64.
 - ID: 0x0005
 
+# API Considerations {#apis}
+
+Some VOPRF APIs specified in this document are fallible. For example, `Finalize()`
```
armfazh, line 1186:
```suggestion
Some VOPRF APIs specified in this document are fallible. For example, `Finalize`
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1169,6 +1181,20 @@ and ristretto255. See {{cryptanalysis}} for related discussion.
...
+and `Evaluate` can fail any element received from the peer fails deserialization.
+The explicit errors generated throughout this specification, along with the
+conditions that lead to each error, are as follows:
+
+- `VerifyError`: VOPRF proof verification failed; {{verifiable-blind}}.
+- `DeserializeError`: Public element or secret scalar deserialization failure; {{pog}}.
```
armfazh, line 1192: can we remove the public and secret adverbs?
Better to generalize to `group elements and scalars`. 

```suggestion
- `DeserializeError`: Group element or scalar deserialization failure; {{pog}}.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1340,6 +1366,33 @@ tolerate discrete logarithm security of lower than 128 bits, we
 recommend only implementing ciphersuites with IDs: 0x0002, 0x0004, and
 0x0005.
 
+## Element and Scalar Validation {#input-validation}
+
+The DeserializeElement function converts an arbitrary byte array to a
```
armfazh, line 1371: Something like this would be better: 
`The DeserializeElement function recovers a group element from its byte representation after serialization.`
similarly for `DeserializeScalar`

armfazh (changes requested):

draft-irtf-cfrg-voprf.md
chris-wood: Hmm, I'm not sure. I think what we have here is accurate. (The person calling this has no idea that the byte array came from the output of SerializeElement, for example. It could be garbage!)

armfazh: as is, it describes also the signature of a hash to group function. 
The issue is with the `convert` word, `recover` could be more accurate.

armfazh: as is, it says that the function necessarily returns a group element from whatever byte string is passed.

chris-wood: Both conversion and recovery are fallible, so I don't really see a difference. But I changed it to recovery anyway. :)

armfazh (approved):

PR #256, opened 2021-07-02
Move definitions around. Clarify DKP implementation and blinding mode for test vectors.

chris-wood:
cc @bytemare, @kevinlewi

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -1529,8 +1529,9 @@ Test vectors with batch size B > 1 have inputs separated by a comma
 "Input", "Blind", "BlindedElement", "EvaluationElement", and
 "Output" fields.
 
-Base and verifiable mode use multiplicative and additive blinding as
-described in {{base-client}} and {{verifiable-client}}, respectively.
+Base mode uses multiplicative blinding while verifiable mode 
+uses additive blinding, as described in {{base-client}} and 
+{{verifiable-client}}, respectively.
 
 The server key material, `pkSm` and `skSm`, are listed under the mode for
 each ciphersuite. Both `pkSm` and `skSm` are the serialized values of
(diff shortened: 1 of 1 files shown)


alxdavids (approved):

draft-irtf-cfrg-voprf.md
```diff
@@ -1537,10 +1529,21 @@ Test vectors with batch size B > 1 have inputs separated by a comma
 "Input", "Blind", "BlindedElement", "EvaluationElement", and
 "Output" fields.
 
+Base and verifiable mode use multiplicative and additive blinding as
+described in {{base-client}} and {{verifiable-client}}, respectively.
```
armfazh, line 1533: Although it's verbose, it is easier to understand.

```suggestion
Base mode uses multiplicative blinding while verifiable mode 
uses additive blinding, as described in {{base-client}} and 
{{verifiable-client}}, respectively.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -551,6 +535,14 @@ In this section, we detail the APIs available on the client and server
 - contextString, a domain separation tag taken from the client or server
   context.
 
+The data type `ClientInput` is an opaque byte string of arbitrary length
+no larger than 2^13 octets. `Proof` is a concatenated sequence of two
+`SerializedScalar` values, as shown below.
+
+~~~
+SerializedScalar Proof[2*Ns];
```
armfazh, line 543: Either `opaque Proof[2*Ns]` or `SerializedScalar Proof[2]` is correct.

armfazh (changes requested):
two comments

draft-irtf-cfrg-voprf.md
chris-wood: Sadly, the latter isn’t. The number denotes the byte count, not element count. We were misusing this all along.

armfazh: It's correct as is. Reference: https://www.rfc-editor.org/rfc/rfc8446.html#section-3.4

armfazh (approved):

PR #265, opened 2021-08-18
Flatten metadata and cast it as auxiliary "info".

chris-wood:
We need to do a pass over the introduction and abstract, as we're no longer building a OPRF/VOPRF, but rather a POPRF/VPOPRF(?).

cc @armfazh, @alxdavids


armfazh:
ACK, on my list for reviewing.

PR #275, opened 2021-09-13
Swap order 

### LINKED ISSUE #274: Out of order context strings: sage code vs. spec
The way the context is constructed in the sage code for the server's evaluate() function does not match the spec.

Sage code:

https://github.com/cfrg/draft-irtf-cfrg-voprf/blob/master/poc/oprf.sage#L87-L89

But the spec says:

  context = "Context-" || contextString ||
            I2OSP(len(info), 2) || info

I believe the sage code should be changed to read as:

        context_DST = _as_bytes("Context-") + self.context_string
        context = context_DST + I2OSP(len(info), 2) + info

The same change also needs to be done to all of the functions that use the "Context-" string: Unblind/Finalize, also for the verifiable version.

chris-wood:
Closes #274.


kevinlewi:
Thanks! Verified that this matches with ristretto255-sha512 in base mode.

PR #303, opened 2022-01-02
add a c implementation to readme

aldenml:
For your consideration (relevant files):

https://github.com/aldenml/ecc/blob/master/src/oprf.h
https://github.com/aldenml/ecc/blob/master/src/oprf.c
https://github.com/aldenml/ecc/blob/master/test/test_oprf.c


chris-wood (approved):
🎉  nice work!

Issue #306, opened 2022-01-11
Specify key derivation


chris-wood:
The current spec punts on key derivation. In particular, it assumes some mechanism for deriving keys. To promote an EdDSA-like interface, wherein "private keys" are random seeds used to _derive_ actual scalars, it would be useful to specify the key generation mechanism. This might look something like the following:

~~~
def DeriveKeyPair(seed, mode, suiteID, info):
  contextString = "VOPRF08-" || 
                  I2OSP(mode, 1) || I2OSP(suiteID, 2) ||
                  I2OSP(len(info), 2) || info
  skS = GG.HashToScalar(seed, DST = "DeriveKeyPair" || contextString)
  if skS == 0:
    raise DeriveKeyPairError
  pkS = ScalarBaseMult(skS)
  return skS, pkS
~~~

One could use this to derive keys for the OPRF, VOPRF, or POPRF protocol variants. For the POPRF case, `info` is repeated here and in the Evaluate interface, but this is not too dissimilar from HPKE, which admits application-specific information in the key schedule and at the AEAD level. Applications can choose how to expose `info` accordingly.

armfazh:
(thinking out loud): when deriving keys based on the mode+info, we do this because it is expected to separate the key spaces, so --for example-- we ensure that generating keys for OPRF are likely not the same than those for VOPRF.

However, there is no way to determine whether a pair of keys (sk,pk) should be used in a particular protocol. In other words, any valid pair (sk,pk) can be used for any protocol variant. Is this something we need to prevent?

chris-wood:
> Is this something we need to prevent?

It's a good question, and something @tomrist and I discussed. There's (at least) two ways we can approach the question:

1. Assume applications are capable of generating keeping keys separate. That is, treat key generation or derivation as something _external_ to the protocol.
2. Assume applications are incapable of keeping keys separate, and enforce key diversification _within_ the protocol.

Option (2) is similar to EdDSA, which treats "private keys" as random seeds and then derives scalars and public group elements from those seeds within the signing APIs. An application that then reuses the same seed across protocol modes would be safe, as the underlying scalars used for evaluation would necessarily be different.

[...]

PR #304, opened 2022-01-07
Major refactor to focus on three protocol variants: OPRF, VOPRF, and POPRF

chris-wood:
This refactor is predicated on the observation that 2HashDH and 3HashSDHI are equivalent for the non-verifiable case, and that public input is only really relevant for the verifiable case. The end result is three protocol variants in the spec: a base OPRF mode that is not variable, a 2HashDH verifiable mode without public input, and a 3HashDH-SDHI variant that is verifiable with public input. 

I think this is a good compromise given all the considerations involved -- threshold VOPRF, huzzah! -- and in the end yields a more understandable document. I also think the new structure should be easier to follow. 

[Rendered version of this PR](https://cfrg.github.io/draft-irtf-cfrg-voprf/caw/all-three/draft-irtf-cfrg-voprf.html).

Note: test vectors have not yet been updated. I will do that in a followup series of commits once this gets some approvals.

cc @kevinlewi, @bytemare, @hugokraw, @tomrist

PR #308, opened 2022-01-13
DeriveKeyPair and serialization improvements

chris-wood:
Based on #304. Addresses #306.

[Rendered version of this PR.](https://cfrg.github.io/draft-irtf-cfrg-voprf/caw/dkp-and-serialize/draft-irtf-cfrg-voprf.html)


draft-irtf-cfrg-voprf.md
```diff
@@ -809,73 +843,67 @@ parameters are assumed to exist:
...
+  blindedElement = blind * P
 
   return blind, blindedElement
 ~~~
 
-Clients store `blind` locally, and send `blindedElement` to the server for evaluation.
-Upon receipt, servers evaluate `blindedElement` using the `Evaluate` function
-described below.
+Clients store `blind` locally, and sends `blindedElement` to the server for evaluation.
+Upon receipt, servers processs `blindedElement` using the `Evaluate` function described
```
bytemare, line 876:
```suggestion
Upon receipt, servers process `blindedElement` using the `Evaluate` function described
```

bytemare (changes requested):

draft-irtf-cfrg-voprf.md
```diff
@@ -762,6 +760,42 @@ def CreateContextString(mode, suiteID):
...
+
+def DeriveKeyPair(seed, info):
+  contextString = CreateContextString(mode, suiteID)
+  deriveInput = seed || I2OSP(len(info), 2) || info
+  counter = 0
+  skS = 0
+  while skS == 0:
+    if counter > 255:
+      raise DeriveKeyPairError
+    skS = GG.HashToScalar(deriveInput || I2OSP(counter, 1),
```
bytemare, line 792: `counter` is bound between 0 and 255, so technically an unsigned 8-bit integer, a byte. Does it make sense to explicitly I2OSP it to one byte?

draft-irtf-cfrg-voprf.md
```diff
@@ -1156,8 +1167,15 @@ and ristretto255. See {{cryptanalysis}} for related discussion.
...
+recommendations, explicit error treatment, and public input representation for the
+POPRF protocol variant.
+
+## External Interface Recommendations
+
+The protocol functions in {{online}} are specified in terms of prime-order group
+Elements and Scalars. However, applications can treat these as internal functions,
+and instead expose interfaces that operate in terms of wire format messages.
 
 ## Error Considerations
```
bytemare, line 1180: This part of the section

> Some OPRF variants specified in this document have fallible operations. For example, `Finalize`
and `Evaluate` can fail if any element received from the peer fails deserialization.

might need to be updated, since Finalize doesn't deserialize anymore

bytemare (changes requested):

draft-irtf-cfrg-voprf.md
```diff
@@ -345,7 +345,8 @@ The following functions and notation are used throughout the document.
 For serialization, all data structure descriptions use TLS notation {{RFC8446, Section 3}}.
 
 All algorithms and procedures described in this document are laid out
-in a Python-like pseudocode.
+in a Python-like pseudocode. The data types `PrivateInput` and `PublicInput`
+are opaque byte strings of arbitrary length no larger than 2^13 octets.
```
armfazh, line 349: we can leverage the TLS syntax notation to express these types.

draft-irtf-cfrg-voprf.md
```diff
@@ -685,35 +685,37 @@ This interaction is shown below.
...
-the `Evaluate` function is the same key as it used to produce the server's public key.
-This proof does not reveal the server's private key to the client. This interaction
-is shown below.
+the `Evaluate` function is the same key as it used to produce the server's public key, `pkS`,
+which the client receives as input to the protocol. This proof does not reveal the server's
+private key to the client. This interaction is shown below.
 
 ~~~
-    Client(pkS)                                     Server(skS, pkS)
+    Client(pkS)            <---- pkS ------                 Server(skS)
```
armfazh, line 693: not really agree with this arrow. it was implicit that the client obtained server's key by any method, not precisely server sending its key to client.
e.g. Client could get key by pre-installation.

armfazh: otoh, does the server require to use a certificate that proves ownership of the pair of keys? (*I was told the protocol is PKI-free)

draft-irtf-cfrg-voprf.md
```diff
@@ -747,13 +750,8 @@ Additionally, each protocol variant is instantiated with a ciphersuite,
 or suite. Each ciphersuite is identified with a two-byte value, referred
 to as `suiteID`; see {{ciphersuites}} for the registry of initial values.
```
armfazh, line 751:
```suggestion
to as `ID`; see {{ciphersuites}} for the registry of initial values.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -762,6 +760,42 @@ def CreateContextString(mode, suiteID):
 
 [[RFC editor: please change "VOPRF08" to "RFCXXXX", where XXXX is the final number, here and elsewhere before publication.]]
 
+## Key Generation and Context Setup {#offline}
+
+In the offline setup phase, both the client and server create a context used
+for executing the online phase of the protocol after agreeing on a mode and
+ciphersuite value suiteID. The server key pair (`skS`, `pkS`) is generated
+using the following function, which accepts a randmly generated seed of length
```
armfazh, line 768:
```suggestion
using the following function, which accepts a randomly generated seed of length
```

draft-irtf-cfrg-voprf.md
```diff
@@ -884,7 +912,6 @@ Output:
 Errors: DeserializeError
 
 def Finalize(input, blind, evaluatedElement):
-  Z = GG.DeserializeElement(evaluatedElement)
   N = blind^(-1) * Z
```
armfazh, line 915: Z is not defined

draft-irtf-cfrg-voprf.md
```diff
@@ -946,12 +967,10 @@ Output:
 Errors: DeserializeError, VerifyError
```
armfazh, line 967:
```suggestion
Errors: VerifyError
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1023,16 +1038,12 @@ Errors: DeserializeError, VerifyError
...
   context = "Info" || I2OSP(len(info), 2) || info
   m = GG.HashToScalar(context)
-
-  R = GG.DeserializeElement(blindedElement)
-  Z = GG.DeserializeElement(evaluatedElement)
-
   T = ScalarBaseMult(m)
   U = T + pkS
-  if VerifyProof(G, U, Z, R, proof) == false:
+  if VerifyProof(G, U, evaluatedElement, blindedElement, proof) == false:
```
armfazh, line 1043: as U is a multiple of G, 
so blinded is a multiple of evaluated element, (double check)

draft-irtf-cfrg-voprf.md
```diff
@@ -1156,8 +1167,15 @@ and ristretto255. See {{cryptanalysis}} for related discussion.
...
-treatment and public input representation for the POPRF protocol variant.
+This section describes considerations for applications, including external interface
+recommendations, explicit error treatment, and public input representation for the
+POPRF protocol variant.
+
+## External Interface Recommendations
+
+The protocol functions in {{online}} are specified in terms of prime-order group
+Elements and Scalars. However, applications can treat these as internal functions,
+and instead expose interfaces that operate in terms of wire format messages.
```
armfazh, line 1178: this means that applications can define its own serialization for objects? or applications must comply with the serialization recommended in this doc? not clear to me.

armfazh (changes requested):

draft-irtf-cfrg-voprf.md
chris-wood: I don't think that's appropriate. These are just byte strings, not length prefixed byte strings or anything.

draft-irtf-cfrg-voprf.md
chris-wood: This simply means the client gets the public key from the server, somehow. That can be preinstallation.

draft-irtf-cfrg-voprf.md
chris-wood: It means that how applications choose to expose the external API is up to them. Does that clarify?

armfazh: last question, does the wire format is no more a MUST in order to be compliant with this draft?

draft-irtf-cfrg-voprf.md
chris-wood: are you saying this is incorrect as is?

draft-irtf-cfrg-voprf.md
chris-wood: Applications that use the OPRF define how to send protocol messages. They can use the wire formats (serialized forms) defined here or something else.

draft-irtf-cfrg-voprf.md
armfazh: I double-checked, it's fine because of the inverse of `t`

draft-irtf-cfrg-voprf.md
chris-wood: Yeah, I think so. We don't say the type of counter is an unsigned byte, so this makes it explicit.

chris-wood:
Thanks for the reviews! I'll update based on the latest comments and then update test vectors, too. (This will change how keys are derived, so the vectors will change.)

chris-wood:
This branch is incredibly screwed up. Blowing it away and starting anew.

Issue #307, opened 2022-01-12
Unhandled zero scalars and identity elements


daxpedda:
This is a list of functions that contain group operations that might produce zero scalars or identity elements. I don't actually know if this is a problem or not in some of the cases, I'm going to leave that to actual cryptographers to figure out.

- [`GenerateProof`](https://cfrg.github.io/draft-irtf-cfrg-voprf/draft-irtf-cfrg-voprf.html#section-2.2.1-2) can return a zero scalar because of `s = (r - c * k) mod p`.

- [`ComputeComposites`](https://cfrg.github.io/draft-irtf-cfrg-voprf/draft-irtf-cfrg-voprf.html#section-2.2.2-4) and [`ComputeCompositesFast`](https://cfrg.github.io/draft-irtf-cfrg-voprf/draft-irtf-cfrg-voprf.html#section-2.2.1-4) use `HashToScalar`, which can return a zero scalar, which is then used in `M = di * Cs[i] + M` and `Z = di * Ds[i] + Z`. Adding `M`, prevents producing an identity point, so it might not be a problem.

- POPRF's [`Finalize`](https://cfrg.github.io/draft-irtf-cfrg-voprf/draft-irtf-cfrg-voprf.html#section-3.3.3-4) uses `HashToScalar`, which can return a zero scalar, which is then used in `T = ScalarBaseMult(m)`, which would produce the identity point. Afterwards, in `U = T + pkS`, adding a non-identity point would prevent `U` from becoming an identity point, so it might not be a problem.

chris-wood:
@armfazh did some analysis and concluded that the only place where we need to be concerned with distinguished scalars is in the `Blind()` step. Fortunately, `Blind` uses `RandomScalar` to produce a scalar, which returns a _non-zero_ scalar. Thus, the only way for `Blind` to produce the identity element is for `HashToGroup` to produce the identity element, and since that is entirely determined by the input to `Blind`, the only thing the function can do is raise an error.

hugokraw:
Chris asked me to take a look at this. The 2HashDH OPRF needs to specify the bind factor to be different than 0 since otherwise the de-blinding would not work. Pathological cases such as getting 0 when a random field value is chosen or getting the identity element when hashing into a curve (and possibly other cases like this) need to be handled as a severe error. Indeed, since the probability that this will happen by chance is essentially zero, such an output indicates a serious bug. 

[...]

PR #310, opened 2022-01-18
Add DeriveKeyPair and separate serialization code from main protocol routines

### LINKED ISSUE #306: Specify key derivation

chris-wood:
Closes #306.

[Rendered version of this PR](https://cfrg.github.io/draft-irtf-cfrg-voprf/caw/dkp-with-serialization/draft-irtf-cfrg-voprf.html)

cc @bytemare, @kevinlewi, @tomrist


draft-irtf-cfrg-voprf.md
```diff
@@ -762,6 +762,42 @@ def CreateContextString(mode, suiteID):
...
+  contextString = CreateContextString(mode, suiteID)
+  deriveInput = seed || I2OSP(len(info), 2) || info
+  counter = 0
+  skS = 0
+  while skS == 0:
+    if counter > 255:
+      raise DeriveKeyPairError
+    skS = GG.HashToScalar(deriveInput || I2OSP(counter, 1),
+                          DST = "DeriveKeyPair" || contextString)
+    counter = counter + 1
```
daxpedda, line 796: Is there interest in making this function infallible?

IANA cryptographer, but could `HashToScalar` be changed to always return a non-zero scalar by using the reduction mod n-1 plus one method. My assumption here is actually that `HashToScalar` returning zero scalars is actually a problem in other functions.

@armfazh mentions here https://github.com/cfrg/draft-irtf-cfrg-opaque/issues/298#issuecomment-1006936162 that `HashToScalar` shouldn't be subject to change, may I ask why?

I made a list of problems here https://github.com/cfrg/draft-irtf-cfrg-voprf/issues/307 that are affected by `HashToScalar` returning a zero scalar.

chris-wood: The probably that this fails is negligible, so it's pretty reasonable for implementations to panic if this occurs. In any case, we'll address fallible functions separately, not in this PR.

armfazh: > @armfazh mentions here [cfrg/draft-irtf-cfrg-opaque#298 (comment)](https://github.com/cfrg/draft-irtf-cfrg-opaque/issues/298#issuecomment-1006936162) that `HashToScalar` shouldn't be subject to change, may I ask why?

We would like the hash function behaves as a random oracle to the entire domain, in this case the whole field including zero.
This is similar to a regular hash function, for example, we cannot require that SHA-256 must not return 0.

armfazh:
We can avoid the counter overflows by using an iterated hashing instead.
what about something like this?
```python
1. T = 000000
2. contextString = CreateContextString(mode, suiteID)
3. deriveInput = seed || I2OSP(len(info), 2) || info
4. MSG = deriveInput || T
5. DST = "DeriveKeyPair" || contextString
6. skS = GG.HashToScalar(MSG, DST)  
7. if skS == 0
8.    T = Hash(T)
9.    return to Step 4.
A. pks = ScalarBaseMult(skS)
B. return (skS,pkS).
```

chris-wood:
> We can avoid the counter overflows by using an iterated hashing instead.

Probability of counter overflow is negligible, so I don't think we need to be overly clever with the construction here. It seems best to match key generation routines used for other algorithms like [HPKE](https://cfrg.github.io/draft-irtf-cfrg-hpke/draft-irtf-cfrg-hpke.html#name-derivekeypair). I suggest we stick with what's here in this PR, and file this iterated hash variant as an issue to possibly do later on.

draft-irtf-cfrg-voprf.md
daxpedda: > In any case, we'll address fallible functions separately, not in this PR.

Sounds good!

armfazh (approved):

PR #311, opened 2022-01-22
Use a separate info string for the DeriveKeyPair info

chris-wood:
Previously, the same `info` string was used for the POPRF evaluation and DeriveKeyPair function. Separating them is an improvement.

cc @bytemare


bytemare:
We're good! 👍

Issue #322, opened 2022-02-03
Update reference implementation to match pseudocode


chris-wood:
Naming of parameters and arguments has fallen behind the pseudocode description, even though the two match functionally. We should update the reference implementation to make things easier to line up. While we're in there, we should also provide some documentation for folks that want to use the reference implementation in other specifications.

PR #325, opened 2022-02-04
Fix batch test vector generation

### LINKED ISSUE #323: Test vectors for POPRF: batching?
Should the test vectors for POPRF include batch sizes > 1, since they support batch evaluation?

### LINKED ISSUE #324: Remove extraneous reference to "Context-" string
Here: https://github.com/cfrg/draft-irtf-cfrg-voprf/blame/main/draft-irtf-cfrg-voprf.md#L345 this sentence makes a little less sense now that "Context-" doesn't actually appear anywhere in the protocol anymore. Perhaps we could consider renaming this example to something that does appear, like "Finalize"?

chris-wood:
Small bug in the test generation code. Thanks, @kevinlewi!

This change trickles down and affects other test vectors due to the way we fix the RNG seed for tests. (It depends on the order in which `random` is invoked, which is silly and something we should fix. I'll do that in #322.)

Closes #323.
Closes #324.


bytemare:
lgtm ✅

PR #338, opened 2022-04-29
Removes OS2IP, and fixes some words.

armfazh:
This PR solves the following comments from [review](https://mailarchive.ietf.org/arch/msg/crypto-panel/QXC3drh1hs-XGkXyjwyCj_pFNNI/): 

>   2.  1.3 on page 7: Better to separate I2SOP and OS2IP, because the current description might be taken to suggest that I2SOP converts a byte array to a non-negative integer, which according to RFC 8017 is just the opposite, i.e., converts a non-negative integer to a byte array.
>   4.  2 on page 8: Redundant long at the end of this sentence "A cryptographic hash function whose output length is Nh bytes long".
>   6.  2.1 on page 9 says "... multiplicative inverse of input Scalar s modulo ...". This is the first time modulo operation is mentioned.

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -335,9 +335,9 @@ The following functions and notation are used throughout the document.
 - For any object `x`, we write `len(x)` to denote its length in bytes.
 - For two byte arrays `x` and `y`, write `x || y` to denote their
   concatenation.
-- I2OSP and OS2IP: Convert a byte array to and from a non-negative
-  integer as described in {{!RFC8017}}. Note that these functions
-  operate on byte arrays in big-endian byte order.
+- I2OSP(x, xLen): Converts a non-negative integer `x` into a byte array
+  of specified length `xLen` as described in {{!RFC8017}}. Note that
+  this function returns a byte array in big-endian byte order.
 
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-voprf.md
```diff
@@ -335,9 +335,9 @@ The following functions and notation are used throughout the document.
 - For any object `x`, we write `len(x)` to denote its length in bytes.
 - For two byte arrays `x` and `y`, write `x || y` to denote their
   concatenation.
-- I2OSP and OS2IP: Convert a byte array to and from a non-negative
-  integer as described in {{!RFC8017}}. Note that these functions
-  operate on byte arrays in big-endian byte order.
+- I2OSP(x, xLen): Converts a non-negative integer `x` into a byte array
```
chris-wood, line 338: Do we not use `OS2IP`?

chris-wood (approved):
LGTM assuming we don't actually using `OS2IP`.

draft-irtf-cfrg-voprf.md
armfazh: I checked it, it is not used, could you confirm too?

chris-wood: Confirmed.

PR #339, opened 2022-04-29
Gathers all labels in a single section.

armfazh:
This PR addresses the comment from [review](https://mailarchive.ietf.org/arch/msg/crypto-panel/QXC3drh1hs-XGkXyjwyCj_pFNNI/)

>  3.  1.3 on page 8 says "String values such as "Finalize" are ASCII string literals": It would be good to have all the strings mentioned here not just an example.


chris-wood (changes requested):
I think this is a reduction in readability. I think the right way to address this comment is merely to list a couple more examples beyond "Finalize".

armfazh:
> I think this is a reduction in readability. I think the right way to address this comment is merely to list a couple more examples beyond "Finalize".

Then, I will decline this one and open a different PR.

PR #343, opened 2022-05-24
Edit input validation text.

### LINKED ISSUE #342: Deserialization error
From [review](https://mailarchive.ietf.org/arch/msg/crypto-panel/QXC3drh1hs-XGkXyjwyCj_pFNNI/)

> 2.1 on page 9 says "This function can raise a DeserializeError if deserialization fails or A is the identity element of the group": This implies that if I serialize and then serialize the identity element, I will get an error!

armfazh:
Edit input validation text.

Changes:
- Move the non-identity requirement from DeserializeElement to Input Validation section.
- Remove SerializedElement and SerializedScalar types, since they are not used.
- Rename validation subsections.
- Minor nits.

Note: usually crypto libraries already implement a `DeserializeElement` functionality (following [SECG Sec 2.3.4](https://www.secg.org/sec1-v2.pdf)), however they do not fail if the identity element is decoded. Hence the check for non-identity must be done explicitly after deserialization. 

Fixes #342

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -1210,8 +1210,8 @@ See {{cryptanalysis}} for related discussion.
 
 ## Input Validation {#input-validation}
 
-Since messages are serialized before transmission between client and server, 
-deserialization is followed by input validation to prevent malformed or invalid
+Since messages are serialized before transmission between client and server,
+deserialization is followed by input validation to prevent malformed or
 invalid inputs from being used in the protocol.
 The DeserializeElement and DeserializeScalar functions instantiated for a
 particular prime-order group corresponding to a ciphersuite MUST adhere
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-voprf.md
```diff
@@ -1215,31 +1210,34 @@ See {{cryptanalysis}} for related discussion.
 
 ## Input Validation {#input-validation}
 
+Since messages between server and clients are usually sent as an opaque
+byte array, a validation process is required to detect malformed or
+invalid inputs of the protocol.
```
chris-wood, line 1215:
```suggestion
Since messages are serialized before transmission between client and server, 
deserialization is followed by input validation to prevent malformed or invalid
invalid inputs from being used in the protocol.
```

chris-wood (approved):
LGTM with one nit.

PR #348, opened 2022-06-28
Update to v10, update test vectors

chris-wood:
cc @kevinlewi and @bytemare, could you please verify vectors if convenient?


bytemare:
Works for [VOPRF ](https://github.com/bytemare/voprf) and [OPAQUE ](https://github.com/bytemare/opaque)as well 👍

PR #355, opened 2022-09-02
Fix comments by JH (part 2)

armfazh:
Together with #354 , this PR addresses all the comments by JH [posted here](https://mailarchive.ietf.org/arch/msg/cfrg/rHVKNpdHx8jWhwqpvlvEKV4pm-4/).

The best way to review this PR is going commit by commit.

## Comments Addressed

> "Verifiable POPRF": the verifiability property is not optional in the
draft, but the fact that it is sometimes called "verifiable POPRF" hints
that it is. Why not avoid any confusion by calling it VPOPRF throughout
the document?

> Section 1.3 introduces PrivateInput and PublicInput as the exact same
data types. What distinguishes them? The terms Private and Public hint
at something, but it seems never spelled out in the document.

> In general the distinction between "Input" and "Parameters" in the
function descriptions confused me. What are the semantics here? First I
thought Parameters are implementation-wide and fixed objects, like the
Group, but then the OPRF key skS appears as a Parameter of
BlindEvaluate. Since the document also hints that keys can be updated
during the run of the protocol ("key rotation"), skS does not seem to be
a long-term parameter to me. Sometimes both the terms Input and
Parameter are mixed, for example when "PublicInput contextString"
appears under Parameters. Another example is in the flow diagram of
POPRF, where both Inputs (info) and Parameters (pkS,skS) appear in
parantheses after a party's name.
Overall it would be good to clarify why the function descriptions
distinguish between "Inputs" and "Parameters", as otherwise implementors
might interpret their own meaning to these terms.

> Explain the list notation []

> Section 3, protocol overviews: skS and pkS appear as input to the
parties but are not used in the protocols. E.g., it should be
BlindEvaluate(skS,blindedElement) instead of BlindEvaluate(blindedElement).

> "client and server augment the pkS and skS, respectively, using the
info value": this could be modified to explain the (otherwise
unexplained) "tweakedKey" term that appears in the protocol overview.
Maybe "client and server tweak their pkS and skS, respectively, using
the info value"?

> If I would implement this, I would wonder about Input "opaque
seed[Ns]". Ns is an integer, and this seed appears to be a list of Ns
opaque values. In the text it says "randomly generated seed". The seeds
suggested in the test vector section do not appear very random to me ;)
Not sure if this is underspecified or just common knowledge.

> "Recall that servers may batch multiple client inputs to
BlindEvaluate." - Maybe one more sentence how this is done? Just
repeated execution of BlindEvaluate and concatenating the
evaluatedElements in the message to the client?

> "certain public inputs that map to invalid public keys for server
evaluation." I could not parse "public key for server evaluation" here.

> I could not make any sense of 5.2, but again, not an implementor...

> Partial obliviousness: I don't understand this definition. Everything
but the last sentence are properties of any OPRF, as described already
in the other properties. The sentence "Both client and server learn the
public input (info)" does not seem to have any meaning, since info is
input to both parties and hence they of course learn it. The partial
obliviousness property rather seems to enforce usage of info in an
evaluation.
I was also surprised that partial obliviousness is equivalent to
unlinkability. Are OPRFs linkable, in the sense of linkability described
in the draft?

> In reference [SJKS17] typos in Jareckiy, Krawczykz

**No change** 

> Still protocol overviews: I find the term "evaluatedElement" computed
by the server a bit misleading, as it sounds like the server is learning
the evaluation. On the other hand I don't have a better idea.
BlindedEvaluatedElement would indicate more precisely what happens, but
is probably too long.

We remain using `evaluatedElement` for the output of the server computation.

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -1471,6 +1471,11 @@ In other words, an attacker with infinite computing power cannot recover any
 information about the client's private input x from an invocation of the
 protocol.
 
+Essentially, input secrecy tells us that, even if the server learns
+the client's private input x at some point in the future, then the server will
+not be able to link any particular PRF evaluation to x. This property is
+also known as unlinkability {{DGSTV18}}.
+
 For the VOPRF and POPRF protocol variants, there is an additional
 security property:
 
@@ -1488,15 +1493,13 @@ commitment as a public key.
 
 Finally, the POPRF variant also has the following security property:
 
-- Partial obliviousness: The server must learn nothing about the client's
-  private input or the output of the function. In addition, the client must
-  learn nothing about the server's private key. Both client and server learn
-  the public input (info).
+- Partial obliviousness: The client and server must be able to perform the
+  PRF on client's private input and public input. The server must learn nothing
+  about the client's private input or the output of the function. In addition,
+  the client must learn nothing about the server's private key.
 
-Essentially, partial obliviousness tells us that, even if the server learns
-the client's private input x at some point in the future, then the server will
-not be able to link any particular POPRF evaluation to x. This property is
-also known as unlinkability {{DGSTV18}}.
+This property becomes useful when dealing with key management operations such as
+the rotation of server's keys.
 
 ## Security Assumptions {#cryptanalysis}
 
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -256,6 +256,9 @@ The following functions and notation are used throughout the document.
 - I2OSP(x, xLen): Converts a non-negative integer `x` into a byte array
   of specified length `xLen` as described in {{!RFC8017}}. Note that
   this function returns a byte array in big-endian byte order.
+- The notation `T U[N]` refers to an array called U containing N items of type
+  T. The type `opaque` means one single byte of uninterpreted data. Items of
+  the array are zero-indexed and referred as `U[j]` such that 0 <= j < N.
 
 All algorithms and procedures described in this document are laid out
 in a Python-like pseudocode. Each function takes a set of inputs and parameters
@@ -1606,7 +1609,7 @@ byte string. The fields of each test vector are described below.
 - "ProofRandomScalar": The random scalar `r` computed in `GenerateProof()`, a
   serialized `Scalar` of `Ns` bytes long. Only present for VOPRF and POPRF
   test vectors.
-- "Output": The protocol output, a byte string of length `Nh` bytes.
+- "Output": The protocol output, an opaque byte string of length `Nh` bytes.
 
 Test vectors with batch size B > 1 have inputs separated by a comma
 ",". Applicable test vectors will have B different values for the
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -1463,8 +1463,8 @@ For the VOPRF and POPRF protocol variants, there is an additional
 security property:
 
 - Verifiable: The client must only complete execution of the protocol if
-  it can successfully assert that the POPRF output it computes is
-  correct. This is taken with respect to the POPRF key held by the
+  it can successfully assert that the output it computes is
+  correct. This is taken with respect to the private key held by the
   server.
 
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-voprf.md
```diff
@@ -256,10 +271,19 @@ The following functions and notation are used throughout the document.
...
-in a Python-like pseudocode. The data types `PrivateInput` and `PublicInput`
-are opaque byte strings of arbitrary length no larger than 2^13 octets.
+in a Python-like pseudocode. Each function takes a set of inputs and parameters
+and produces a set of output values. Parameters become constant values once the
+protocol variant and the ciphersuite are fixed.
+
+The `PrivateInput` data type refers to information that must be kept in
+secrecy, and the `PublicInput` data type indicates that it is not required to
+keep the information secret. These data types are opaque byte strings of
+arbitrary length no larger than 2^13 octets.
```
chris-wood, line 286:
```suggestion
The `PrivateInput` data type refers to inputs that are known only to the client
in the protocol, whereas the `PublicInput` data type refers to inputs that are 
known to both client and server in the protocol. Both `PrivateInput` and
`PublicInput` are opaque byte strings of arbitrary length no larger than 2^13 octets.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -536,8 +560,9 @@ def VerifyProof(A, B, C, D, proof):
             "Challenge"
 
   expectedC = G.HashToScalar(h2Input)
+  verified = expectedC == c
```
chris-wood, line 563:
```suggestion
  verified = (expectedC == c)
```

draft-irtf-cfrg-voprf.md
```diff
@@ -837,23 +864,22 @@ below.
...
+def BlindEvaluate(skS, blindedElement):
   evaluatedElement = skS * blindedElement
   return evaluatedElement
 ~~~
 
 Servers send the output `evaluatedElement` to clients for processing.
-Recall that servers may batch multiple client inputs to `BlindEvaluate`.
+Recall that servers may process multiple client inputs by applying the
+`BlindEvaluate` function to each `blindedElement` received, and returning an
+array with the corresponding `evaluatedElement`s.
```
chris-wood, line 882:
```suggestion
array with the corresponding `evaluatedElement` values.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1455,12 +1483,17 @@ In other words, an attacker with infinite computing power cannot recover any
 information about the client's private input x from an invocation of the
 protocol.
 
+Essentially, input secrecy tells us that, even if the server learns
```
chris-wood, line 1486:
```suggestion
Essentially, input secrecy is the property that, even if the server learns
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1455,12 +1483,17 @@ In other words, an attacker with infinite computing power cannot recover any
...
+the client's private input x at some point in the future, then the server will
+not be able to link any particular PRF evaluation to x. This property is
+also known as unlinkability {{DGSTV18}}.
```
chris-wood, line 1489:
```suggestion
the client's private input x at some point in the future, the server cannot
link any particular PRF evaluation to x. This property is also known as unlinkability {{DGSTV18}}.
```

draft-irtf-cfrg-voprf.md
```diff
@@ -1549,7 +1580,7 @@ P-384 (used by 0x0004), or P-521 (used by 0x0005).
 
 ## Domain Separation {#domain-separation}
 
-Applications SHOULD construct input to the protocol to provide domain
+Applications MUST construct input to the protocol to provide domain
```
chris-wood, line 1583:
```suggestion
Applications SHOULD construct input to the protocol to provide domain
```

chris-wood: A MUST isn't enforceable.

chris-wood (changes requested):
Looks good modulo a few changes. I don't quite agree with making skS a parameter, but if it moves us forward faster, let's do it!

chris-wood (approved):

PR #363, opened 2022-10-15
Addresses Chris Patton comments.

### LINKED ISSUE #362: RGLC Review: Chris P.: Editorial
- [x] (nit) s.2.1: In the list of group member functions, each item is prefixed by "A member function of Group that" except for the first three. I think you could remove these prefixes.
- [ ] s.2.2: "noninteractive" -> "non-interactive"
- [x] s.3: "In this section, we define three protocol variants referred as the OPRF, VOPRF, and POPRF modes with the following properties." -> "referred to as" 
- [x] (nit) s.3.1: Table 1: "Identifiers for OPRF modes" -> "identifiers for protocol types" (VOPRF, POPRF, OPRF are not "modes of OPRF")
- [x] s.3.3: "in which case the application MUST abort the protocol with a DeserializeError failure": Elsewhere (s.2.1) you write "raise a DeserializeError". Consistent verbiage would be helpful I think. (The latter seems better for this draft.)
- [ ] s.4: "each group specifies an integer order that is used in reducing integer values to a member of the corresponding scalar field": "modulus" is clearer than "integer order" here.
- [x] s.4.3: "In the elliptic curve setting, this deterministically maps inputs x (as byte arrays) to uniformly chosen points on the curve": Here a variable "x" is named but not used. I think this sentence would be perfectly readable without "x".
- [ ] s.5.4: "Functionally, the VOPRF and POPRF variants differ in that the POPRF variant admits public input, whereas the VOPRF variant does not." The same statement applies equally to OPRF and POPRF. Maybe write "(V)OPRF" here instead?
- [x] s.6: "This section discusses the cryptographic security of our protocol": "protocol" -> "protocols" or "protocol variants". Also (nit): The word "cryptographic" doesn't add much here, in my opinion.
- [x] s.6.1: "A consequence of showing that a function is pseudorandom, is that it is necessarily non-malleable": Remove the "," in this sentence.
- [ ] s.6.1: "Verifiable: The client must only complete execution of the protocol if it can successfully assert that the output it computes is correct. This is taken with respect to the private key held by the server." Do you mean "public key" here? (Just checking.)
- [x] s.6.2.1: "[[JKK14](https://www.ietf.org/archive/id/draft-irtf-cfrg-voprf-14.html#JKK14)] proves these properties for one instance": I read "JKK14" as "Jarecki, Kiayias, and Krawczyk, 2014", hence I would write "prove" instead of "proves" here.
- [x] s.6.2.2: "The construction is identical to 3HashSDHI, except that this design can optionally perform multiple POPRF evaluations in one go, ...": "in one go" is an English colloquialism, and thus may not be appropriate for non-native English speaking audiences.
- [x] s.6.2.3: "As a result of this class of attack, ...": Hmm, should it be "class of attacks"?

armfazh:
Addresses Chris Patton (@cjpatton) comments posted at https://mailarchive.ietf.org/arch/msg/cfrg/l4QDx8HbE0egEsVcW9qtX_AnN7Y/

fixes #362

### DIFF
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -1517,7 +1517,7 @@ admit public input.
 
 # Security Considerations {#sec}
 
-This section discusses the security of OPRF protocols, along
+This section discusses the security of the protocols defined in this specification, along
 with some suggestions and trade-offs that arise from the implementation
 of the protocol variants in this document. Note that the syntax of the POPRF
 variant is different from that of the OPRF and VOPRF variants since it
--- a/draft-irtf-cfrg-voprf.md
+++ b/draft-irtf-cfrg-voprf.md
@@ -354,7 +354,7 @@ prime-order group.
 - HashToScalar(x): Deterministically maps
   an array of bytes `x` to an element in GF(p). This function is optionally
   parameterized by a DST; see {{ciphersuites}}. Security properties of this
-  function are described in Section 10.5 of {{!I-D.irtf-cfrg-hash-to-curve}}.
+  function are described in {{!I-D.irtf-cfrg-hash-to-curve, Section 10.5}}.
 - RandomScalar(): Chooses at random a non-zero element in GF(p).
 - ScalarInverse(s): Returns the inverse of input `Scalar` `s` on `GF(p)`.
 - SerializeElement(A): Maps an `Element` `A`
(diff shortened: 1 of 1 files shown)


draft-irtf-cfrg-voprf.md
```diff
@@ -1518,9 +1517,9 @@ admit public input.
 
 # Security Considerations {#sec}
 
-This section discusses the cryptographic security of our protocol, along
+This section discusses the security of OPRF protocols, along
```
chris-wood, line 1520:
```suggestion
This section discusses the security of the protocols defined in this specification, along
```

chris-wood (approved):

PR #373, opened 2022-12-19
replaces HTML by ASCII in equation

armfazh:


armfazh:
looks like the browser disagrees.

https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-voprf#name-rejection-sampling

image

chris-wood:
I fixed this in #374 -- please take a look!

armfazh:
I don't know what is wrong with the changes in this pr, could you please elaborate.
The pro's of these changes is that we don't need to embed HTML codes in the document that is rendered not only in HTML, but PDF and ASCII versions too.

chris-wood:
The problem is that this doesn't render like we'd want it to in HTML. Please take a look at #374 -- this produces the desired output.

armfazh:
declined in favour of PR-374

PR #374, opened 2022-12-20
Fix math rendering

chris-wood:
Alternative to #373 

cc @armfazh


armfazh (commented):
~the problem I see is not with the comparison symbol, but with the HTML tags. 
HTML tags do not render in the PDF & TXT versions.~

Ok, I just checked the artifacts produced in this job, and confirm this solves the issue.

armfazh (approved):

PR #388, opened 2023-09-01
Replacing hash to curve draft with RFC9380.

armfazh:
Referencing to published RFC9380


armfazh:
this will be done by rfc editor

